PR #41, opened 2019-05-20
Reorder clock IDs to match implementation

MarkMcCaskey:
We had  a small bug in our implementation because we assumed that the order listed in the docs was the order of the enumeration starting at 0.  This PR makes that assumption correct


sunfishcode:
Sorry I missed this before; I've now filed #112 which applies this fix to the new phases directory structure.

PR #47, opened 2019-05-30
Add a writeup of current optional imports design

### LINKED ISSUE #36: Weak imports
Wasm doesn't have a "weak" import mechanism, which would allow one to `(import "foo" "bar")` where "bar" may or may not exist in "foo" (or perhaps where "foo" itself may not exist). This would be one way to allow feature detection in APIs.

As one possible solution, it's expected that [the wasm GC proposal](https://github.com/WebAssembly/gc) will eventually give wasm the concept of optional/nullable references, which would permit a kind of weak import where one imports global variables with reference type which can be either references to the desired entities or absent/null. However, GC may not be available in wasm for some time, and even when it is, it may not be implemented in all wasm engines, so it's not entirely obvious whether we should depend on it for something so fundamental to WASI.

Another is to essentially add a weak-import system on top of Wasm. For example, we could use a naming convention, such as a ".weak" suffix, to indicate a weak import, and a ".is_present" suffix on a global variable import to hold a boolean indicating whether a weak import was successfully resolved, like this:

```wasm
  (func $foo (import "wasi_something" "foo.weak"))
  (global $do_we_have_foo (import "wasi_something" "foo.is_present") i32)
```
A similar scheme could enable detection of modules too:
```wasm
   (func $foo (import "wasi_something.weak" "foo"))
   (global $do_we_have_wasi_something (import "wasi_something.is_present" "foo") i32)
```
or even both at the same time:
```wasm
   (func $qux (import "wasi_something.weak" "bar.weak"))
   (global $do_we_have_it (import "wasi_something.is_present" "foo.is_present") i32)
```

Another option would be to have a dynamic API that takes a string, however while we [can do that](https://github.com/WebAssembly/webidl-bindings), passing around strings introduces some complexity and it seems like if have a naming convention for ".weak", we might as well have ".is_present" too.

I'm not attached to any of the details here, I'm mainly just looking to start a discussion.

sbc100:
Document how weak imports are specified and the contents of the new custom section
used for this purpose

Fixes #36


design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+
+```wasm
+(func $statvfs (import "wasi:fs" "statvfs.weak"))
+(global $statvfs_is_present (import "wasm:fs" "statvfs.is_present") i32)
+```
+
+On older systems that don't support the new function `$statvfs_is_present` would
+be set to zero and calling `$statvfs` would result in a trap.
+
+On system that do support the new function `$statvfs_is_present` is set to
```
cjihrig:
```suggestion
On systems that do support the new function `$statvfs_is_present` is set to
```

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+function imports.
+
+Declaring a weak function import
+--------------------------------
+
+Weak function imports are implemened by using two imports for each function.
+The first being the weak function and the second being a WebAssembly i32 global
+to indicate if the function is present at runtime.
+
+For example, if a module called `wasi:fs` added a new `statvfs` function to it
```
cjihrig:
```suggestion
For example, if a module called `wasi:fs` added a new `statvfs` function to its
```

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+API.  For example, if a WASI API evolves to include a new function a program
+might want to continue to run on older system that don't yet support the new
+function.  In this case a weak import mechanism allows the program to run on
+older systems and detect the presence of the function at runtime.
+
+WebAssembly itself does not currently provide a mechanism for weak imports.
+There is some discussion of adding it the spec [^1], and WASI would
+likely use such a feature if/when it is added.
+
+This document describes the mechanism used by WASI to achieve a form a weak
```
cjihrig:
```suggestion
This document describes the mechanism used by WASI to achieve a form of weak
```

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
+Weak Imports
+============
+
+It can be useful for WASI programs to weakly depend certain functions in a WASI
+API.  For example, if a WASI API evolves to include a new function a program
+might want to continue to run on older system that don't yet support the new
```
cjihrig:
```suggestion
might want to continue to run on older systems that don't yet support the new
```

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
+Weak Imports
+============
+
+It can be useful for WASI programs to weakly depend certain functions in a WASI
```
cjihrig:
```suggestion
It can be useful for WASI programs to weakly depend on certain functions in a WASI
```

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+be set to zero and calling `$statvfs` would result in a trap.
+
+On system that do support the new function `$statvfs_is_present` is set to
+non-zero and calling `$statvfs` would work as expected.
+
+Using a weak function import
+----------------------------
+
+In order to use the above weak function import its presence should first be
+tested for.  In C code this would looks something like this:
```
cjihrig:
```suggestion
tested for.  In C code this would look something like this:
```

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+
+```c
+if (wasm_fs.statvfs) {
+  wasm_fs.statvfs(...)
+}
+```
+
+At the WebAssembly level might look like this:
```
cjihrig:
```suggestion
At the WebAssembly level it might look like this:
```

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+```wasm
+(func $statvfs (import "wasi:fs" "statvfs.weak"))
+(global $statvfs_is_present (import "wasm:fs" "statvfs.is_present") i32)
+```
+
+On older systems that don't support the new function `$statvfs_is_present` would
+be set to zero and calling `$statvfs` would result in a trap.
+
+On system that do support the new function `$statvfs_is_present` is set to
+non-zero and calling `$statvfs` would work as expected.
```
alexcrichton: In the spirit of determinism and such, perhaps this could be specified to be a particular nonzero value? (e.g. 1)

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
+Weak Imports
+============
+
+It can be useful for WASI programs to weakly depend on certain functions in a
+WASI API.  For example, if a WASI API evolves to include a new function a
+program might want to continue to run on older systems that don't yet support
+the new function.  In this case a weak import mechanism allows the program to
+run on older systems and detect the presence of the function at runtime.
```
sunfishcode: Another use case worth mentioning here is APIs which aren't applicable in a given embedding, such as filesystem APIs on a device with no filesystem.

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+Weak function imports are implemened by using two imports for each function.
+The first being the weak function and the second being a WebAssembly i32 global
+to indicate if the function is present at runtime.
+
+For example, if a module called `wasi:fs` added a new `statvfs` function to its
+interface a program could import this new function weakly in the following way:
+
+```wasm
+(func $statvfs (import "wasi:fs" "statvfs.weak"))
+(global $statvfs_is_present (import "wasm:fs" "statvfs.is_present") i32)
```
sunfishcode: What if we put the ".weak" and ".is_present" in the module name, rather than the field name? This is a little less intuitive, however it would address the concern about intruding on what would otherwise be the user's or compiler's namespace. `(import "wasi:fs.weak" "statvfs")` and `(import "wasi:fs.is_present" "statvfs")`?

design/weak-imports.md
```diff
@@ -0,0 +1,60 @@
...
+the new function.  In this case a weak import mechanism allows the program to
+run on older systems and detect the presence of the function at runtime.
+
+WebAssembly itself does not currently provide a mechanism for weak imports.
+There is some discussion of adding it the spec [^1], and WASI would
+likely use such a feature if/when it is added.
+
+This document describes the mechanism used by WASI to achieve a form of weak
+import on top of the existing primitives.  Currently this is only defined for
+function imports.
```
sunfishcode: We should also mention that this feature is not related to weak GC references, since we've already had one question about this.

design/weak-imports.md
sbc100: Interesting.  Kind of nasty but see you point.   What do others think of this idea?

binji: Another option would be to define a reserved character to use instead of `.`. It's unlikely any system would want to use `\0` for its symbol names, for example.

pchickey: Is there some existing encoding that we could reuse here, e.g. url encoding?

sbc100: Using null bytes in strings seems a like asking for trouble.

dumblob: If any character (combination), then definitely something visible in ASCII.

sbc100:
I've updated this PR, going back the custom section design, as discussed in the last meeting.

sbc100:
Update with name change to "optional"

sbc100:
@sunfishcode ptal

sunfishcode:
Looks like all the feedback has been addressed! If anyone has further feedback, please file issues in the issue tracker.

Issue #24, opened 2019-04-28
WASI Libraries


CryZe:
So I'm using wasm with wasmer + cranelift for a kind of scripting system for my application. So users can write little WebAssembly modules to write plugins. Over the weekend I've looked into integrating WASI to give them access to the operating system in a limited fashion (I'm planning to only let them read files and print to stdout / stderr initially). However it seems like so far (at least from Rust, not sure if this is an overall limitation) you can really only create WASI binaries, not libraries. The way my plugin system works is that my application expects certain exports to be there and I call into them, while providing WASI imports + other plugin specific imports to the wasm files. However at least the way it currently is in Rust and the wasi libc sysroot, WASI really only works for the duration of a main function that needs to be there. Calling the wasi functions from anywhere else (through the Rust std) segfaults because libpreopen isn't initialized (anymore). I'm not sure how much this is a design limitation or a just a bug, but overall I'd like my main application to have control over which functions to call. With a main function, the WebAssembly module takes control over the control flow, which isn't great for such an "extension" kind of use case.

ericsink:
I bumped into this problem as well, although from C instead of Rust, so I'm not sure how of much of this applies to your situation.

Yes, it appears that _startup() initializes libpreopen before it calls main(), so if I don't want _startup() or main(), I have a problem.

I'm assuming that the answer is that if I compile with

    -Wl,--no-entry

as an argument to clang, then I am supposed to take responsibility for initializing libpreopen myself.

There seem to be functions in <wasi/libc.h> and <wasi/libc-find-relpath.h> to be used for that purpose.

But I haven't actually tried using them yet.

PoignardAzur:
Wait, is preopen necessary for calling WASI "syscall" functions? I thought it was only used as a polyfill for standard library calls using absolute paths?

If you want to make a plugin system (eg, for a text editor) that passes directory capabilities to its plugin and doesn't allow absolute paths, do you still need preopen? Or the standard library, for that matter?

CryZe:
The main reason I'm even using WASI is to be able to open and read from files in the plugins. Not using the standard library for that would probably also mean that I just don't need WASI. But I feel like that defeats the purpose of WASI if you can't even use it in such use cases.

I feel like the standard library should either lazily initialize preopen if you don't go through main, or there should be __wasi_init() and __wasi_deinit() exports in case you use it as a library, which the runtime can use instead of main to initialize preopen and whatever else the WASI module needs.

sbc100:
[...]

PR #48, opened 2019-05-30
Add application-abi describing the current unstable WASI application ABI

sbc100:
This is that start of a document regarding what we expect from a valid was WASI application.    Not ready to land yet. 

Does this kind of thing belong here?  If not then where?

Perhaps I should include here documentation of the current status quo (i.e. _start function etc) along with the forward looking plans?

See https://github.com/WebAssembly/WASI/issues/24 and https://github.com/WebAssembly/WASI/issues/19

### DIFF
--- a/design/application-abi.md
+++ b/design/application-abi.md
@@ -6,8 +6,8 @@ are also certain expectations that the WASI runtime places on an application
 that wishes to be portable across WASI implementations.
 
 This document describes how a conforming WASI application is expected to behave
-in terms of lifecycle (startup, shutdown, etc) and any exports the that
-it is expected to include.
+in terms of lifecycle (startup, shutdown, etc) and any exports it is expected to
+include.
 
 Current Unstable ABI
 --------------------
@@ -18,12 +18,12 @@ application:
 - `_start` - the program entry point
 - `memory` - linear memory used by the program
 
-The `_start` start export must be WebAssembly function and will be used as the
-program entry point.  This is the default name used by `lld` when linking
-WebAssembly modules.  The embedder is expected to call this function once the
-module is instantiated.
+The `_start` export must be WebAssembly function and will be used as the program
+entry point.  This is the default name used by `lld` when linking WebAssembly
+modules.  The embedder is expected to call this function once the module is
+instantiated.
 
-Many of current WASI unstable APIs require a sharing of linear memory between
+Many of the current WASI unstable APIs require sharing of linear memory between
 the application and the embedder.  In order to use any such APIs the WASI module
 is expected to export its linear memory under the name `memory`.
 
--- a/design/application-abi.md
+++ b/design/application-abi.md
@@ -6,27 +6,32 @@ are also certain expectations that the WASI runtime places on an application
 that wishes to be portable across WASI implementations.
 
 This document describes how a conforming WASI application is expected to behave
-in terms of lifecycle (startup, shutdown, etc) and imports and exports the that
+in terms of lifecycle (startup, shutdown, etc) and any exports the that
 it is expected to include.
 
-Lifecycle
----------
+Current Unstable ABI
+--------------------
 
-A WASI program may contain any number exports, but the embedded attributes
-specific meaning the following optional function exports:
+The current WASI unstable ABI specifies only two exports from a WASI
+application:
 
-- `__wasi_init` - Called after WebAssemembly instantiation but before any other
-  functions.
-- `__wasi_main` - Runs the programs main entry points.  May be omitted, for
-  example, in the case of a library.
-- `__wasi_term` - The logical inverse of `__wasi_init`.  Optionally called
-  before module destruction.  No other functions within program will be called
-  after this one.
+- `_start` - the program entry point
+- `memory` - linear memory used by the program
 
-Linear Memory
--------------
+The `_start` start export must be WebAssembly function and will be used as the
+program entry point.  This is the default name used by `lld` when linking
+WebAssembly modules.  The embedder is expected to call this function once the
+module is instantiated.
 
-All WASI programs are expected to share a linear memory with the embedder.  The
-memory can either be imported from the embedder or exports to the embedder.  If
-exported the memory must be named `__wasi_memory`.  If imported the import must
-be named `memory` from a module names `wasi`.
+Many of current WASI unstable APIs require a sharing of linear memory between
+the application and the embedder.  In order to use any such APIs the WASI module
+is expected to export its linear memory under the name `memory`.
+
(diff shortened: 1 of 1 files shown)


design/application-abi.md
```diff
@@ -0,0 +1,32 @@
...
+A WASI program may contain any number exports, but the embedded attributes
+specific meaning the following optional function exports:
+
+- `__wasi_init` - Called after WebAssemembly instantiation but before any other
+  functions.
+- `__wasi_main` - Runs the programs main entry points.  May be omitted, for
+  example, in the case of a library.
+- `__wasi_term` - The logical inverse of `__wasi_init`.  Optionally called
+  before module destruction.  No other functions within program will be called
+  after this one.
```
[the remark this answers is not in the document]
sbc100: I was also thinking along those lines.   I was basing this on the fact that we do currently have "post main" cleanup code in crt1.c:

```
    /* Call atexit functions, destructors, stdio cleanup, etc. */                
    __prepare_for_exit();    
```

However if we switch the Command model than such cleanup can stay where it as and run after main.  So for Commands yes I agree __wasi_term maybe is not needed.

For the case of Reactors would you not also want code like this (e.g. to flush stdout and run exit handler) in the case of a normal/ideal shutdown?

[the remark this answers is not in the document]
sunfishcode: I think this comes down to how Reactors will terminate. I imagine they'll terminate by calling `proc_exit` (or [whatever that ends up as](https://github.com/WebAssembly/WASI/issues/26)). In that case, they can do `__prepare_for_exit` or whatever other cleanup they need to do themselves at that time.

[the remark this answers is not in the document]
sbc100: You don't think there is any need for user/embedder of the Reactor to inform the Reactor when its don't using it?  Otherwise how does it know when it should prepare for its exit?

Another way of putting it:  `proc_exit` is something that a WASI application (the Reactor) would have when its exiting.     How do I as the user/embedder tell the reactor I'm done with it?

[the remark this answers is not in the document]
sunfishcode: I'm not sure about this, but one theory is that in a Reactor, there is no need for such a thing. Reactors shouldn't have data sitting around in a stdout buffer between events if they really care about it getting delivered, because in many contexts, there won't be a chance for a shutdown notification -- it'll be power-off-the-IoT-device, STONITH-the-server, WIFI-just-disappeared, abort-the-blockchain-transaction, or other things. I don't completely believe in this theory yet, but it would simplify some things.

I was going to add closer-the-browser-tab, but in that case, there is an `onbeforeunload` callback. If we want to expose the `onbeforeunload` callback to wasm via `__wasi_term`, we may want to adjust the wording to account for the fact that unloading can be cancelled and restarted, so perhaps `__wasi_term` may mean that no further functions will be called.

So, I'm ok adding `__wasi_term` here for now, and we an continue to think about this.

design/application-abi.md
```diff
@@ -0,0 +1,32 @@
...
+
+Linear Memory
+-------------
+
+All WASI programs are expected to share a linear memory with the embedder.  The
+memory can either be imported from the embedder or exports to the embedder.  If
+exported the memory must be named `__wasi_memory`.  If imported the import must
```
wanderer: do we need to have the ability to both import and export memory? What would be the usecase for importing memory that isn't covered just by exporting?

sbc100: I was actually surprised to see that the wasi toolchain currently exports memory by default.  I suppose that is because wasm-ld happens to have that is its default.

As a point of reference emscripten module import their memory from the embedder by default. 

If we want to support sharing of memories the importing is the only way to go.  If every module creates and exports it own memory there is no way they can access the same one.

I suppose we could start by documenting the status quo which is to export the memory from the module, and then extend this later as needed?

sunfishcode: It was also a conscious decision to have modules define their own memories -- looking forward to wasm modules compiled from languages which don't use linear memories, or which use multiple linear memories, it seems like it will be awkward for the host to try to provide exports for the things applications will need. The Command/Reactor/Library ontology can help here as well -- Commands and Reactors would (typically) define the main memory, and Libraries would (typically) import it.

sbc100:
I paired down this PR to simply document the current ABI.   Will move the rest to a followup.

sunfishcode:
This looks good to me. We can leave this open a little longer to give people a chance to comment, but with the latest commits, it's now paired down (see @sbc100's comment) to just documenting the status quo, with pointers to ongoing discussions, so I don't expect it's controversial.

design/application-abi.md
```diff
@@ -0,0 +1,37 @@
+WASI Application ABI
+====================
+
+In addition to the APIs defined by the various WASI [modules](modules.md) there
+are also certain expectations that the WASI runtime places on an application
+that wishes to be portable across WASI implementations.
+
+This document describes how a conforming WASI application is expected to behave
+in terms of lifecycle (startup, shutdown, etc) and any exports the that
```
cjihrig:
```suggestion
in terms of lifecycle (startup, shutdown, etc) and any exports
```

design/application-abi.md
```diff
@@ -0,0 +1,37 @@
...
+Current Unstable ABI
+--------------------
+
+The current WASI unstable ABI specifies only two exports from a WASI
+application:
+
+- `_start` - the program entry point
+- `memory` - linear memory used by the program
+
+The `_start` start export must be WebAssembly function and will be used as the
```
cjihrig:
```suggestion
The `_start` export must be a WebAssembly function and will be used as the
```

design/application-abi.md
```diff
@@ -0,0 +1,37 @@
...
+program entry point.  This is the default name used by `lld` when linking
+WebAssembly modules.  The embedder is expected to call this function once the
+module is instantiated.
+
+Many of current WASI unstable APIs require a sharing of linear memory between
```
cjihrig:
```suggestion
Many current WASI unstable APIs require sharing of linear memory between
```

design/application-abi.md
```diff
@@ -0,0 +1,37 @@
...
+in terms of lifecycle (startup, shutdown, etc) and any exports it is expected to
+include.
+
+Current Unstable ABI
+--------------------
+
+The current WASI unstable ABI specifies only two exports from a WASI
+application:
+
+- `_start` - the program entry point
```
xtuc: @lukewagner @guybedford in https://github.com/nodejs/node/pull/27850#issuecomment-499475321 we discussed the issue of running the main on import. However, this assumption seems to be wrong? Nodejs could decide to expose the `_start` function to the user.

PR #69, opened 2019-07-18
add virtualization doc

wanderer:
This PR attempts to sketch out what the overloaded term "virtualization" mean, why we might want it and how it might be accomplished.

### DIFF
--- a/docs/virtualization.md
+++ b/docs/virtualization.md
@@ -1,0 +1,113 @@
+# WatIDL :duck:
+
+WatIDL (`/ˈwädl/`) is an IDL for webassembly. Its goal is to be able to fully describe the import and export capabilities of Webassembly modules while being intuitive to read and reason about. It inherits the Webassembly text format's syntax and attempts to closely follow and fit into the Webassembly ecosystem. WatIDL should used as a simple syntax for describing the interfaces along with WebIDL bindings to describe how the different data structures are formed and received by Wasm modules.
+
+# Differences from WAT
+While WatIDL follow close to WAT there are some differences
+
+## Types
+WatIDL uses the following primitive types
+
+```
+refTypes := "anyRef" | "funcRef"
+intType := "s8" | "s6" | "s32" | "s64" | "u8" | "u16" | "u32" | "u64";
+floatType ::= "f32" | "f64" 
+primType ::= "null" | "bool" | "string" | "data" | intType | floatType | refTypes;
+```
+
+In addition new types can also be created using
+- struct
+- array
+- union
+- enum
+
+The syntax for these follows the [gc proposals](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md) syntax when possible.
+
+```
+(type $drinks
+  (enum
+    $coffee
+    $juice
+    $water
+    $wine
+  )
+)
+
+(type $baz
+  (struct
+    (field $foo $drinks)
+    (field $bar bool)
+  )
+)
+```
+
+## Interface and Functions
+### Imports and Exports
+Instead of having "module"s WatIDL has "interfaces" which has nearly the same syntax as [modules](https://webassembly.github.io/spec/core/text/modules.html). An interfaces "exports" are to be imported by a module using the interface and the interfaces "imports" are to be exported by the module using the interface. Since imports are to be provided by the module using the interface the have no module namespace associated with them.  
+
+```
+(interface $a
+  ;; the module using this interface must export "memory"
+  (import "memory" (memory)) 
+  (global $a (export "a_global") (mut s32) (s32.const -2))
+  (func (export "a_func"))
+)
+```
+
+### Results
+watidl can have multiple results and results can be labeled
+```
+(func (export "multi-vale")
+  (result $result1 u64)
+  (result $result2 f32)
+)
+```
+
+### Object Oriented
+An interface is also a first class type
+
+```
+(interface $b)
+
+(interface $a
+  (func (export "factory")
+    (result $the_result $b)
+  )
+)
+```
+
+### inheritance
+An interface can extend a base interface inheriting its imports, functions and types
+
+```
+(interface $b
+  (func (export "add")
+    (param u64 u64)
+    (result u64)
+  )
+)
+
+(interface $a extends $b
+  (func (export "sub")
+    (param u64 u64)
+    (result u64)
+  )
+)
+```
+
+### Method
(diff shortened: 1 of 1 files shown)


docs/virtualization.md
```diff
@@ -0,0 +1,113 @@
...
+(interface $b)
+
+(interface $a
+  (func (export "factory")
+    (result $the_result $b)
+  )
+)
+```
+
+### inheritance
```
MarkMcCaskey: Is the reason to use inheritance over composition here because of other OOP types defined in the IDL?

Being able to arbitrarily combine interfaces without the restrictions of a hierarchy is something I'm interested in if this grows beyond WASI

wanderer: ohh i really messed this up. I copied the wrong file here!!! :S please move to https://github.com/WebAssembly/WASI/pull/64

sunfishcode:
This document introduces the concepts of "static" and "dynamic" virtualization, however it appears to only discuss "dynamic" virtualization, with techniques to essentially create vtable-like objects which would expose the API functions.

[This paper](http://www.cs.cmu.edu/~aldrich/papers/ecoop17modules.pdf) describes a path which enables virtualizating static APIs dynamically by making modules first-class objects. From what I understand, this would seem to fit wasm fairly well, as it would allow us to continue to use static bindings for many cases, but still virtualize APIs. I'd be interested in what you think of this approach.

lukewagner:
Agreed with @sunfishcode.  I've also discussed the rough idea with @erights and it seemed like it would indeed be possible to base an ocap-style runtime-polymorphic system on top of a lower-level, more-static virtualization mechanism like what's in the paper Dan linked to.  I think it would end up being expressed as a "layered on top of core wasm" spec, just like, and quite complementary to, Interface Types.

Unfortunately, we don't yet have even an explainer-level writeup of how the idea in the paper maps concretely down to wasm.  Personally, I'm interested to help write an initial sketch so that we can get a Stage 0 proposal repo to discuss more, but I need to make some more progress on getting Interface Types written up before starting.  I'd also be happy to collaborate with anyone who wants to start sooner.

docs/virtualization.md
```diff
@@ -0,0 +1,62 @@
+# Virtualization
+
+Virtualization in the context of WASI interfaces is the ability for a Webassembly module to implement a given interface and consumer module to be able to use that implement transparently.
```
revmischa:
```suggestion
Virtualization in the context of WASI interfaces is the ability for a Webassembly module to implement a given interface and for a consumer module to be able to use its implementation transparently. 
```

docs/virtualization.md
```diff
@@ -0,0 +1,62 @@
+# Virtualization
+
+Virtualization in the context of WASI interfaces is the ability for a Webassembly module to implement a given interface and consumer module to be able to use that implement transparently. 
+
+Furthermore there are two classes of virtualization; static and dynamic. Static virtualization is done before the initialization of a Wasm store and dynamic at runtime.
```
revmischa:
```suggestion
Furthermore there are two classes of virtualization: static and dynamic. Static virtualization is done before the initialization of a Wasm store and dynamic is performed at runtime.  
```

docs/virtualization.md
```diff
@@ -0,0 +1,62 @@
...
+
+Furthermore there are two classes of virtualization; static and dynamic. Static virtualization is done before the initialization of a Wasm store and dynamic at runtime.  
+
+## Usecase
+
+Roughly taken from [here](https://pdfs.semanticscholar.org/4cce/9abb177cc58c199e13b82d498f37010c2bfc.pdf). Lets say a user wants to encrypt a text file they are editing. One way this could be accomplished is by having composable services. An encryption module could export the interface for a directory that would encrypt any file written to it. Lets consider how this could be done with in the framework of WASI. At runtime the encryption module would given a root directory file descriptor in which it would write encrypted text to, it would also would return a wrapped file descriptor of a directory  that would be given to the text editor to write in. This means any WASI `fd` related calls that the text editor makes would handled by the encryption module which would then interact base system. Lets look next at a WASI program might accomplish virtualizing the file descriptor interface.
```
revmischa:
```suggestion
Roughly taken from [here](https://pdfs.semanticscholar.org/4cce/9abb177cc58c199e13b82d498f37010c2bfc.pdf). Let's say a user wants to encrypt a text file they are editing. One way this could be accomplished is by having composable services. An encryption module could export the interface for a directory that would encrypt any file written to it. Let's consider how this could be done with in the framework of WASI. At runtime the encryption module would given a root directory file descriptor in which it would write encrypted text to, it would also would return a wrapped file descriptor of a directory that would be given to the text editor to write in. This means any WASI `fd`-related calls that the text editor makes would handled by the encryption module which would then interact with the base system. Let's look next at how a WASI program might accomplish virtualizing the file descriptor interface.
```

docs/virtualization.md
```diff
@@ -0,0 +1,62 @@
...
+(func mk-fd-open-proxy (param (ref $fd) (result (ref $open_func)))
+  (func.bind $open_func (local.get 0) (func.ref $fd-open-proxy))
+)
+```
+
+In this scenario no functions are imported, only types. While this is one possible future we don't currently have this functionality. While it maybe be possible to polyfill things like `func.bind` using tables, other things such as full GC would more complicated and may create overhead.
+
+### Dynamic Dispatching - Without GC 
+
+Assuming the function reference proposal, another way to implement dynamic dispatching currently would be to create a primitive polyfill for structs. An `fd` would then be represented as a stuct who's fields contained all the associated functions
```
revmischa:
```suggestion
Assuming the function reference proposal, another way to implement dynamic dispatching currently would be to create a primitive polyfill for structs. An `fd` would then be represented as a struct whose fields contain all the associated functions.
```

sunfishcode:
Virtualization in WASI is now described in the [Design Principles](https://github.com/WebAssembly/WASI/blob/master/docs/DesignPrinciples.md) document which has now landed, in the section on [interposition](https://github.com/WebAssembly/WASI/blob/master/docs/DesignPrinciples.md#interposition) (meaning "virtualization" in the sense used here), which lays out a vision. In short, the "verbs" are imported functions, and the "nouns" are handle values. This describes the "more static" version described by Luke [above](https://github.com/WebAssembly/WASI/pull/69#issuecomment-535549118).

If there are any questions or concerns about this approach, or anything that's unclear, feel free to file issues or bring them to the WASI Subgroup meetings.

Issue #51, opened 2019-05-31
__wasi_prestat_t documentation is missing


wanderer:
the description of `__wasi_prestat_t` from [`__wasi_fd_prestat_get`](https://github.com/WebAssembly/WASI/blob/master/design/WASI-core.md#__wasi_fd_prestat_get) seems to missing.  Its defined [here](https://github.com/CraneStation/wasmtime/blob/3e94af81df7c34b2e2cfd78482b740b80a2ec1fe/wasmtime-wasi/sandboxed-system-primitives/include/wasmtime_ssp.h#L310) in wasmtime but I'm not sure why  `__wasi_preopentype_t pr_type;` can't just be a `__wasi_filetype_t`

sunfishcode:
This was fixed in c760e95483c357dd814984bba85c2be1c97c3252.

`pr_type` isn't a `__wasi_filetype_t` because in the future we may have more kinds of preopen types, like [preopened clocks](https://github.com/WebAssembly/WASI/issues/118) or other things. We could stand to have [more preopen types than we do now](https://github.com/WebAssembly/WASI/issues/91) though.

PR #74, opened 2019-08-15
A wasi_unstable description based on module types.

sunfishcode:
This sketches out a description of the current wasi_unstable API, using a syntax which anticipates the "module types" syntax [recently proposed to the CG](https://github.com/WebAssembly/meetings/blob/master/2019/CG-08-06.md#discuss-new-proposal-that-introduces-types-for-modules-and-instances-and-text-format-thereof-as-initially-discussed-in-design1289), though it does use a few extensions which can be easily lowered, and which mostly anticipate the [Interface Types](https://github.com/WebAssembly/webidl-bindings/issues/47#issuecomment-519717553) proposal.

This is derived from [the WatIDL proposal](https://github.com/WebAssembly/WASI/pull/64), though it differs in a few ways:
 - It doesn't yet do anything special for capabilities, for now. File descriptors are just integers, for now. But the idea is that we could add OCAP concepts later.

 - It uses a new `(flags ...)` construct in place of structs of Bool fields, to describe flags values. This allows us to explicitly declare the underlying wasm type for flags values.

 - Types used by the API are declared outside of the moduletype, because we're using wat syntax as much as possible, and there, `(type ...)` inside a module declares an entry in the type section, which isn't what's intended here.

The extensions to module types are:
 - It uses `string`, `array`, and `{s,u}{8,16,32,64}` types, which are expected to be provided in some form by the Interface Types proposal.  For now though, they can all be lowered in straightforward ways, by assuming that `string` is always just UTF-8.

 - It adds `typename` declarations, which can name any wasm type, and of the extra types mentioned in the previous item, or `struct`, `enum`, or `flags`.

 - It declares functions with multiple return values, which for now will need to be lowered to output (pointer) parameters.

### DIFF
--- a/design/wasi_unstable/wasi_unstable.wati
+++ b/design/wasi_unstable/wasi_unstable.wati
@@ -1,0 +1,1206 @@
+;; The wasi_unstable module type.
+;;
+;; Some content here is derived from [CloudABI](https://github.com/NuxiNL/cloudabi).
+
+(typename $size_t u32)
+(typename $string_array (array string))
+
+;; Non-negative file size or length of a region within a file.
+(typename $filesize_t u64)
+
+;; Timestamp in nanoseconds.
+(typename $timestamp_t u64)
+
+;; Identifiers for clocks.
+(typename $clockid_t
+  (enum u32
+    ;; The store-wide monotonic clock, which is defined as a clock measuring
+    ;; real time, whose value cannot be adjusted and which cannot have negative
+    ;; clock jumps. The epoch of this clock is undefined. The absolute time
+    ;; value of this clock therefore has no meaning.
+    $CLOCK_MONOTONIC
+    ;; The CPU-time clock associated with the current process.
+    $CLOCK_PROCESS_CPUTIME_ID
+    ;; The clock measuring real time. Time value zero corresponds with
+    ;; 1970-01-01T00:00:00Z.
+    $CLOCK_REALTIME
+    ;; The CPU-time clock associated with the current thread.
+    $CLOCK_THREAD_CPUTIME_ID
+  )
+)
+
+;; Error codes returned by functions.
+;; Not all of these error codes are returned by the functions provided by this
+;; API; some are used in higher-level library layers, and others are provided
+;; merely for alignment with POSIX.
+(typename $errno_t
+  (enum u16
+    ;; No error occurred. System call completed successfully.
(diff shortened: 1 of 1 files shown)


design/wasi_unstable/wasi_unstable.wati
```diff
@@ -0,0 +1,1206 @@
...
+  )
+)
+
+;; File descriptor rights, determining which actions may be performed.
+(typename $rights_t
+  (flags u64
+    ;; The right to invoke __wasi_fd_datasync().
+    ;;
+    ;; If __WASI_RIGHT_PATH_OPEN is set, includes the right to invoke
+    ;` __wasi_path_open() with __WASI_FDFLAG_DSYNC.
```
steveeJ:
```suggestion
    ;; __wasi_path_open() with __WASI_FDFLAG_DSYNC.
```

design/wasi_unstable/wasi_unstable.witx
```diff
@@ -0,0 +1,525 @@
...
+    ;; The number of bytes read.
+    (result $nread $size_t)
+  )
+
+  ;; Return a description of the given preopened file descriptor.
+  (@interface func (export "fd_prestat_get")
+    (param $fd $fd_t)
+    (result $error $errno_t)
+    ;; The buffer where the description is stored.
+    (result $buf $prestat_t)
```
cjihrig: It looks like `prestat_t` is not defined (related to https://github.com/WebAssembly/WASI/issues/51).

sunfishcode: Good spot! I've now added it.

pchickey:
There are a handful of fixes in https://github.com/fastly/lucet/pull/275/files#diff-1ac8557cf1c286d69d13ccaa4b21d291 that get these files to "validate", at least as far as lucet-idl is concerned. Most of the changes were either typo fixes, or changing the ordering of `typename`s so that all types are defined before use.

We don't necessarily have to enforce that names are defined before use, but it does make implementation a bit easier, and it has precedent in wasm.

sunfishcode:
@pchickey Thanks! I've now made an attempt to merge your changes into the PR here.

The wat format does allow '$' names to be used before defined in the file, so we should eventually support that. That said, it's not urgent, and I've incorporated your reordering changes in the PR here as well for now.

PR #90, opened 2019-09-13
Add witx crate, to parse and validate witx

pchickey:
As a follow on to #74, I'm submitting the `witx` crate, adapted from the Lucet project (https://github.com/fastly/lucet).

`witx` parses and validates a witx file into the representation given in `tools/witx/src/ast.rs`. The validation rules are written in `src/validation.rs`. The crate produces a binary, `witx`, which validates a witx file, or reports an error if it fails to validate.

`witx` uses stable rust, and has only two crate dependencies - `failure`, a very-nearly-standard error reporting library, and `clap`, a popular argument parsing library. (`clap` and the executable could be separated out into a separate crate if necessary to drop that dependency).

This PR also fixes one minor syntax error in the existing witx files. It is already useful!

After this is merged, I will make a PR to add Azure Pipelines integration to ensure that this code builds without errors, and validates all checked-in witx files, as a condition to merge each PR.


sunfishcode:
Looks good! Since we discussed this in the last meeting, let's go ahead and merge, and we can continue to iterate with PRs and issues from here.

PR #92, opened 2019-09-18
Fix the name of `wasi_ephemeral_preview0` and update tests.

sunfishcode:
`wasi_ephemeral_preview0` is meant to be versioned, so that it can be
superceded by new versions in the future.

Also, update the witx testsuite to test all the current witx files.


sbc100:
Should we add circleci to run the tests?  (No necessarily as part of this PR)

PR #94, opened 2019-09-19
add azure-pipelines files, mostly copied from cranelift/wasmtime

pchickey:


.azure-pipelines.yml
```diff
@@ -0,0 +1,124 @@
...
+      workingDirectory: tools/witx
+      displayName: Cargo build
+
+    - bash: cargo test --all
+      workingDirectory: tools/witx
+      displayName: Cargo test
+      env:
+        RUST_BACKTRACE: 1
+
+- job: Build
```
alexcrichton: Most of the goop here (weird env vars, centos, libstdc++, etc) is only related to the binary releases of the `wasmtime` binary and associated tools. Is it intended though that the witx tool is available as a binary through this repo? If not this can probably all largely be dleted, if so though we can probably simplify the Linux build by quite a lot since `witx` doesn't depend on C++ (unlike wasmtime)

sunfishcode:
@alexcrichton Unfortunately Github Actions is stil in Beta, and the WebAssembly organization isn't currently registered to participate in the Beta program. Perhaps we should go with Azure Pipelines for now, and switch to Github Actions when it's released?

pchickey:
I saw a tweet from Nat that access to the actions beta is now granted instantly on requesting it

.azure-pipelines.yml
pchickey: The `witx` executable is useful to distribute, since it validates a `.witx` document.

And yeah, lets drop all the C++ stuff as well. I didn't actually know what all of that was there for.

alexcrichton: Ok cool, in that case if we're building releases then I think we'll want to switch the Linux release to using the `x86_64-unknown-linux-musl` target to statically link everything. That way we don't have to deal with centos and can use a modern build environment. The Windows/OSX bits would need to stay, but this isn't likely to be a super widely used executable anyway so none of the weird build config is honestly all that important to keep.

sbc100: Yeah, I would hope we can avoid all that docker + sudo + static linking complexity.  Building against a standard debian/stable (of ubuntu LTS perhaps) image should be good enough to produce a widely runnable binary.

.azure-pipelines.yml
```diff
@@ -0,0 +1,124 @@
...
+        toolchain: stable
+      mac-stable:
+        imageName: 'macos-10.14'
+        toolchain: stable
+      mac-beta:
+        imageName: 'macos-10.14'
+        toolchain: beta
+      mac-nightly:
+        imageName: 'macos-10.14'
+        toolchain: nightly
```
sbc100: Probably just the 3 stable builds should be enough, no?

pchickey:
Closing in favor of letting @alexcrichton setup Github Actions instead.

.azure-pipelines.yml
tschneidereit: @sbc100 it's not really, unfortunately: we encountered issues with distros more than a few years old, and setting up the container and all that was the way to solve those. With this setup, the produced binaries run on distros all the way back to ~2007 or so.

sbc100: Building against a debian/stable or debian/stable - 1 should be enough to get compat with the vast vast majority of actual linux users.  At least that has been our experience with with chrome, native-client-sdk and emsdk.  This is even more true for developer machines which is that target of these binaries.

sunfishcode:
Github Actions is now enabled for this repository, so @alexcrichton if you're able to set up Github Actions rules here, that'd be great!

.azure-pipelines.yml
tschneidereit: Oh, I entirely agree that that's sufficient for this use case, as developers are really unlikely to run anything that old for actual development work :) For Wasmtime the situation is a bit different, but that's neither here nor there.

PR #96, opened 2019-09-25
witx crate: improve public API

pchickey:
* Friendly public API to `Document` and `Module` - expose a getter and an iterator for each kind of definition in each, rather than expose the internal Vec<Definition>/HashMap<Id, Entry> datastructures.
* Implement `PartialEq` and `Eq` for all AST nodes. The Document and Module implementations are implemented manually, and are not sensitive to the order of definitions.
* Make a standalone `parse(&str) -> Result<Document,...>` function available. This uses `MockFs` under the hood.
* Factor out IO to a separate module in the crate. Add variants of error reporting that looks up the location of the error using `WitxIo`
* Add a `Render` trait that converts AST nodes back into s-expressions. A separate SExpr enum is used because it needs to own `String`s, rather than have refs (`&'a str`) and `Location` throughout. Implement `Display` to render SExprs to text. This allows the validated AST to be serialized to a single string or file.

### DIFF
--- a/tools/witx/src/ast.rs
+++ b/tools/witx/src/ast.rs
@@ -18,8 +18,41 @@ impl Id {
 
 #[derive(Debug, Clone)]
 pub struct Document {
-    pub definitions: Vec<Definition>,
-    pub entries: HashMap<Id, Entry>,
+    definitions: Vec<Definition>,
+    entries: HashMap<Id, Entry>,
+}
+
+impl Document {
+    pub(crate) fn new(definitions: Vec<Definition>, entries: HashMap<Id, Entry>) -> Self {
+        Document {
+            definitions,
+            entries,
+        }
+    }
+    pub fn datatype(&self, name: &Id) -> Option<Rc<Datatype>> {
+        self.entries.get(name).and_then(|e| match e {
+            Entry::Datatype(d) => Some(d.upgrade().expect("always possible to upgrade entry")),
+            _ => None,
+        })
+    }
+    pub fn datatypes<'a>(&'a self) -> impl Iterator<Item = Rc<Datatype>> + 'a {
+        self.definitions.iter().filter_map(|d| match d {
+            Definition::Datatype(d) => Some(d.clone()),
+            _ => None,
+        })
+    }
+    pub fn module(&self, name: &Id) -> Option<Rc<Module>> {
+        self.entries.get(&name).and_then(|e| match e {
+            Entry::Module(d) => Some(d.upgrade().expect("always possible to upgrade entry")),
+            _ => None,
+        })
+    }
+    pub fn modules<'a>(&'a self) -> impl Iterator<Item = Rc<Module>> + 'a {
+        self.definitions.iter().filter_map(|d| match d {
+            Definition::Module(d) => Some(d.clone()),
+            _ => None,
+        })
+    }
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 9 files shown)


tools/witx/src/render.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+}
+
+impl Render for IntRepr {
+    fn to_sexpr(&self) -> SExpr {
+        match self {
+            IntRepr::I8 => SExpr::word("u8"),
+            IntRepr::I16 => SExpr::word("u16"),
+            IntRepr::I32 => SExpr::word("u32"),
+            IntRepr::I64 => SExpr::word("u64"),
```
sunfishcode: Since we're now round-tripping `IntRepr` from "u" type to "u" type, it seems like we should rename `IntRepr::I8` etc. to `IntRepr::U8` etc.

tools/witx/src/io.rs
```diff
@@ -0,0 +1,88 @@
+use crate::WitxError;
+use std::collections::HashMap;
+use std::fs::{read_to_string, File};
+use std::io::{BufRead, BufReader, Error, ErrorKind};
+use std::path::{Path, PathBuf};
+
+pub trait WitxIo {
+    fn fgets(&self, path: &Path) -> Result<String, WitxError>;
```
sunfishcode: This would benefit from a doc comment, along the lines of "Read the whole contents of a file into a string".

tools/witx/src/io.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn fget_line(&self, path: &Path, line_num: usize) -> Result<String, WitxError>;
```
sunfishcode: Ditto, possibly mentioning that this is used for error reporting.

tools/witx/src/ast.rs
```diff
@@ -122,10 +183,76 @@ pub struct UnionVariant {
...
+    pub fn funcs<'a>(&'a self) -> impl Iterator<Item = Rc<InterfaceFunc>> + 'a {
+        self.definitions.iter().filter_map(|d| match d {
+            ModuleDefinition::Func(d) => Some(d.clone()),
+            _ => None,
+        })
+    }
 }
 
+impl PartialEq for Module {
+    fn eq(&self, rhs: &Module) -> bool {
```
sunfishcode: Should this function test for the `name` fields being equal too? If not, it'd be good to say why not in a comment.

tools/witx/src/lib.rs
```diff
@@ -17,13 +21,14 @@ pub use ast::{
     ModuleDefinition, ModuleEntry, ModuleImport, ModuleImportVariant, StructDatatype, StructMember,
     UnionDatatype, UnionVariant,
 };
+pub use io::{Filesystem, MockFs, WitxIo};
 pub use lexer::LexError;
 pub use parser::{DeclSyntax, ParseError};
+pub use render::{Render, SExpr as RenderSExpr};
```
sunfishcode: Is `RenderSExpr` used anywhere? Renaming things in imports like this has its uses, but I also find it confusing, so I'm looking for what's motivating it here :-).

tools/witx/src/ast.rs
```diff
@@ -122,10 +183,76 @@ pub struct UnionVariant {
...
+        }
+        for f in self.funcs() {
+            if let Some(f_rhs) = rhs.func(&f.name) {
+                if f != f_rhs {
+                    return false;
+                }
+            } else {
+                return false;
+            }
+        }
```
sunfishcode: Are these two loops the same as doing `self.entries == rhs.entries`? If not, could you add a comment about what's different?

tools/witx/src/lib.rs
```diff
@@ -33,6 +38,14 @@ pub fn load<P: AsRef<Path>>(path: P) -> Result<Document, WitxError> {
     validate_document(&parsed_decls).map_err(WitxError::Validation)
 }
 
+pub fn parse(source: &str) -> Result<Document, WitxError> {
```
sunfishcode: I assume this only works on strings that don't contain `(use ...)` directives? If so, that'd be good to mention in a comment.

tools/witx/src/render.rs
pchickey: Agree, will change this

tools/witx/src/lib.rs
pchickey: It is not used anywhere under that name. When I renamed the import I was thinking it would disambiguate it from `sexpr::SExpr`, which has different types (&strs and locations) in the leaf nodes. But that type isn't publicly exposed by this library, so, perhaps I should just export this one as SExpr. You're correct that it isnt worth any potential confusion.

tools/witx/src/ast.rs
pchickey: that's correct, I just didn't think of that. thank you!

tools/witx/src/ast.rs
sunfishcode: What is the resolution here?

tools/witx/src/ast.rs
sunfishcode: It looks like this isn't updated yet.

pchickey:
Thanks for all the helpful review comments!

pchickey:
Published as witx 0.2.0

PR #97, opened 2019-09-25
Make witx document describing wasi spec available as the `wasi-spec` crate

pchickey:
Based on #96 

Over in Lucet, we need to consume the `witx` description of the WASI spec in several different rust crates. Rather than put each crate in charge of finding and parsing `.witx` files from the filesystem, and and to provide an easy mechanism to use the same version of the spec across the project, I have created the `wasi-spec` crate in this project.

`wasi-spec` has a build.rs step to serialize the witx document into a single file, and uses `include_str!` and `witx::parse` to make the document available in the crate. At the moment it only packages `wasi_unstable_preview0`. We can package the other phases as well if desired.

I'm not sure if packaging the spec as a crate is exactly the right thing to do, but it is useful to me, and might be useful to others as well, so I figured it can live here rather than in the Lucet crate. I am open to feedback about whether this is the right way to package it or the right place for this code to live.


tools/wasi/build.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut f = File::create(witx_path).expect("create file");
+    f.write_all(format!("{}", doc).as_bytes())
+        .expect("write data");
+}
+
+fn main() {
+    for (phase, version) in WASI_VERSIONS {
+        let doc =
+            load_wasi_spec(phase, version).expect(&format!("load wasi {} {}", phase, version));
+        serialize_wasi_spec(phase, version, &doc);
```
sunfishcode: It's not immediately clear what this build.rs program is doing. Why do we need to load and serialize the witx here?

pchickey:
Sorry, this was a bad idea, I didn't realize this won't work unless built from the git repo. The spec .witx files won't be present if this crate is downloaded from crates.io and the build.rs will fail.

PR #100, opened 2019-09-26
wasi_unstable_preview0.witx: fixes to match current implementations

pchickey:
This PR makes changes to `wasi_unstable_preview0.witx` to align with the current implementations (wasi-libc, wasi-common, and the wasi rust crate)

* `environ_sizes_get` returns an `errno_t` as its first result
* `path_open` takes a `fdflags_t` argument.
* `proc_sched_yield` is actually called `sched_yield`

Open question: These changes are seeking to make the witx document descriptive, rather than prescriptive. `wasi_unstable_preview0.witx` is also out of sync with current implementations by using the `wasi_unstable_preview0` module name, where current implementations use `wasi_unstable`.

Should we instead make `phases/old/witx/wasi_unstable.witx` that describes current (pre-`preview0`) implementations? Or should `preview0`?

### DIFF
--- a/phases/ephemeral/witx/wasi_ephemeral_preview0.witx
+++ b/phases/ephemeral/witx/wasi_ephemeral_preview0.witx
@@ -37,6 +37,7 @@
   )
   ;; Return command-line argument data sizes.
   (@interface func (export "environ_sizes_get")
+    (result $error $errno_t)
     ;; The number of arguments.
     (result $argc $size_t)
     ;; The size of the argument string data.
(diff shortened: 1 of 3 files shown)


sunfishcode:
My sense here is that it's early enough that we can go ahead and change `wasi_unstable_preview0`. We won't want to do this in general, but this is a special case since we `wasi_unstable_preview0` didn't go through the `ephemeral` phase where APIs can settle before we promote them.

pchickey:
Agreed, this is a special case because we're trying to get the starting point of the witx spec to match reality (where wasi was specified by the libc's `wasi/core.h`).

Should we also port these fixes into `old/witx/wasi_unstable.witx`?

sunfishcode:
Yes, I think it makes sense to update `old/witx/wasi_unstable.witx` too, so that it reflects what implementations are currently using.

pchickey:
Should the patches be put into `ephemeral` as well, since we want them to be the basis of any future PRs to update the ephemeral version of the spec? My take is that if ephemeral is deciding to depart from the implementations, it should do so explicitly

PR #101, opened 2019-09-28
witx crate: calculate module type signature of an interface function

pchickey:
Also ports in a fix that should have been in #96 .

This PR defines a mapping of interface function types to module types. This mapping reflects the way witx maps to the existing implementations of WASI:
* Array, Data, and String types are passed as a pointer and a length
* Structs and Union types are passed as a pointer
* Pointer and ConstPointer are passed as I32. This is considered passing by value.
* All other types can be passed by value - integer and floats get mapped to the smallest atom that can represent them.
* the first result type is returned by value. validation requires that a pass-by-value datatype is used here.
* subsequent result types are passed as arguments, by reference.

### DIFF
--- a/tools/witx/src/ast.rs
+++ b/tools/witx/src/ast.rs
@@ -137,7 +137,7 @@ pub struct AliasDatatype {
     pub to: DatatypeIdent,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


sunfishcode:
"Module types" is a little confusing, because a [module type](https://github.com/WebAssembly/module-types) is now the type of a module. Perhaps we could call what this PR is introducing "core types"?

pchickey:
I agree the name "module types" is confusing. I'm happy to change it to "core types".

Issue #103, opened 2019-10-01
Change WHENCE constants?


kripken:
It turns out that the `SEEK_CUR, SEEK_SET, SEEK_END` constants are pretty constant between native platforms - enough that in SDL (a very portable codebase), they are simply hardcoded. This caused breakage when I tried to [change emscripten's constants from musl's to wasi's](https://github.com/emscripten-core/emscripten/pull/9555#issuecomment-537100816), since `__WASI_WHENCE_CUR != SEEK_CUR` etc.

We can fix SDL and emscripten etc., but on the other hand, talking with @dschuff we thought maybe it's a good idea to make the wasi constants consistent with existing values. These are arbitrary values anyhow, and wasi is still unstable. This would cause some short-term inconvenience, but maximize compatibility with code in the long term.

Thoughts?

For comparison, errno codes definitely have differences between major platforms, and so there isn't a natural set of values to copy. (We [changed the codes in emscripten to match wasi](https://github.com/emscripten-core/emscripten/pull/9545), and that breaks our ABI but that's ok.)

sunfishcode:
#106 landed, so this is fixed in ephemeral. We don't have a fixed schedule for when ephemeral updates to a versioned module, but I imagine we'll do so once we've batched up a few fixes.

PR #106, opened 2019-10-03
Change whence_t constant values to match pre-existing agreed-upon values.

kripken:
Background: https://github.com/WebAssembly/WASI/issues/103

This reorders the whence constants to match existing values, shared by things like [musl](https://github.com/ifduyue/musl/blob/b4b1e10364c8737a632be61582e05a8d3acf5690/include/stdio.h#L36) and [SDL](https://github.com/SDL-mirror/SDL/blob/c17aa515bdc8fd6efad9c5fbb3f2e2fb777e1bc6/include/SDL_rwops.h#L174).

Without this, using a libc with the old wasi constants will not work in SDL, which assumes those values are fixed. As SDL is a stable and portable codebase, apparently these constants are consistent in practice across different platforms.

We only noticed the SDL issue by chance, so it's very possible there are more such issues out there. By changing the constants we'll minimize code porting issues in the long term. However, in the short term existing wasi runtimes and binaries may disagree.

### DIFF
--- a/phases/ephemeral/docs/wasi_ephemeral_preview.md
+++ b/phases/ephemeral/docs/wasi_ephemeral_preview.md
@@ -2305,6 +2305,10 @@ Used by [`__wasi_fd_seek()`](#fd_seek).
 
 Possible values:
 
+- <a href="#whence.set" name="whence.set"></a>**`__WASI_WHENCE_SET`**
+
+    Seek relative to start-of-file.
+
 - <a href="#whence.cur" name="whence.cur"></a>**`__WASI_WHENCE_CUR`**
 
     Seek relative to current position.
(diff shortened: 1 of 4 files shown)


sbc100:
I agree.   I don't see any reason to diverge from the status quo and the fact the an extensively ported codebase like SDL relies on these values is a good sign that they don't generally vary between platforms. 

We should prepare a parallel patch from wasi-libc.

kripken:
Thanks, changed to ephemeral.

sbc100:
Cloudabi is actually even more strange and uses 1, 2, 3.  No zero at all.   Odd.

sunfishcode:
No objections and this is a minor change, so let's merge!

PR #117, opened 2019-10-16
witx: add handles kind

pchickey:
`handle` is a kind of datatype for foreign references, or capabilities: it is opaque (has no in-memory representation), and is passed as a u32 index.

In ephemeral, `fd_t` is changed to be a handle, rather than a u32.

### DIFF
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -273,8 +273,8 @@
   )
 )
 
-;;; A file descriptor index.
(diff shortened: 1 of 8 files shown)


pchickey:
Need to redo the first commit of this in light of #124 - will try to get it done sometime this week, but its not urgent at the moment.

pchickey:
The parser implementation has been updated

pchickey:
Just rebased on master, filled in docs implementation

PR #124, opened 2019-10-21
Switch to the Rust `wast` crate for parsing `*.witx`

alexcrichton:
This commit switches to the recently-published `wast` crate to parse
`*.witx` file. The `wast` crate is intended to be a way to easily write
recursive descent and composable parsers. It includes built-in parsers
for all of `*.wat` and `*.wast` syntax, but only a few limited ones were
used for `*.witx`, otherwise it's just the idioms that were used for the
`Parse` trait and such!

The hope here is that this can consolidate lexing/handling of
s-expressions and simplify the parsing strategy as well. Eventually it's
intended that `wast` is flexible enough to have built-in parsers for
interface types as well!

The crate was refactored slightly in implementing this change,
prominently:

* The `validate` pass is not baked into the `src/toplevel.rs` parsing.
* The `src/toplevel.rs` parsing was simplified slightly to have only one
  recursive function for parsing includes.
* All `*Syntax` data types now have a `'a` lifetime parameter and are
  consumed immediately on the stack frame while parsing.
* `Location` is no longer manufactured while parsing, but rather it's
  created on-demand while the ast is being constructed. This required a
  small refactoring to have a `DocValidationScope` which carries
  metadata about the current file that was parsed to manufacture
  `Location`.


tools/witx/src/toplevel.rs
```diff
@@ -1,162 +1,144 @@
...
-    use crate::parser::*;
-    use crate::Location;
 
     #[test]
     fn empty() {
-        assert_eq!(
-            parse_witx_with(&Path::new("/a"), &MockFs::new(&[("/a", ";; empty")])).expect("parse"),
-            Vec::new(),
-        );
+        parse_witx_with(&Path::new("/a"), &MockFs::new(&[("/a", ";; empty")])).expect("parse");
```
sunfishcode: Why did the `assert_eq!` part go away here? Should we still be able to assert that `part_witx_with` returns an empty result here?

tools/witx/src/toplevel.rs
```diff
@@ -1,162 +1,144 @@
...
-                &MockFs::new(&[("/a", "(use \"b\")"), ("/b", ";; empty")])
-            )
-            .expect("parse"),
-            Vec::new(),
-        );
+        parse_witx_with(
+            &Path::new("/a"),
+            &MockFs::new(&[("/a", "(use \"b\")"), ("/b", ";; empty")]),
+        )
+        .unwrap();
```
sunfishcode: ditto

tools/witx/src/toplevel.rs
alexcrichton: The interface of this function changed slightly to return a `Document` instead of returning a raw parsed syntax tree, so I ended up changing the tests too. I don't mind preserving the tests though, I could add a separte function to ensure everything is parsed as-is, but I figued it wouldn't add too much over the assertions this has currently.

pchickey: I wouldn't sweat these tests - I want to make changes soon that allow multiple top-levels (because the refactored wasi spec will be spread among many files) and that will change some things about the way `use` works - it is currently a bit fast and loose.

pchickey:
Published as 0.3.1

PR #125, opened 2019-10-21
Remove `subscription_t`'s `clock.identifier` field.

sunfishcode:
As noticed in
https://github.com/rust-lang/rust/pull/65617#issue-330088745,
the `clock.identifier` field in `subscription_t` isn't used for
anything. It came from CloudABI, where it appears to be a holdover from
an earlier API feature which is no longer present.

### DIFF
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -552,8 +552,6 @@
 ;; The contents of a $subscription_t when type is `EVENTTYPE_CLOCK`.
 (typename $subscription_clock_t
   (struct
-    ;; The user-defined unique identifier of the clock.
-    (field $identifier $userdata_t)
     ;; The clock against which to compare the timestamp.
     (field $clock_id $clockid_t)
     ;; The absolute or relative timestamp.


phases/ephemeral/witx/typenames.witx
```diff
@@ -552,8 +552,6 @@
 ;; The contents of a $subscription_t when type is `EVENTTYPE_CLOCK`.
 (typename $subscription_clock_t
   (struct
-    ;; The user-defined unique identifier of the clock.
-    (field $identifier $userdata_t)
```
bjorn3: This may theoretically be useful for debuggers.

sunfishcode: What would it add beyond the `clock_id` field?

bjorn3: I though I read that it was a string. In that case, a debugger could show that string to the user. I think it would be better to design such debugger integrations separately and for all things it makes sense for at a later time.

newpavlov:
@dunnock 
Since it's a breaking change, I think it's worth to synchronize it with other planned Core API changes and release them in a single update.

dunnock:
@newpavlov thanks, it does not seem an entry level stuff then

sunfishcode:
Yes, following the [phases](https://github.com/WebAssembly/WASI/blob/master/phases/README.md) document, this change is happening to ephemeral, which we'll batch up with other changes and snapshot to a versioned interface. As for keeping rust-wasi up to date, I'm working on a way to make that easier; hopefully I'll have the PR ready soon.

PR #127, opened 2019-10-24
Change hard link count to 64-bits.

### LINKED ISSUE #70: Make __wasi_linkcount_t 64-bit
`__wasi_countcount_t` is currently a `uint32_t`. We should change it to `uint64_t` in order to support hypothetical filesystems with very large link counts. For comparison, the [nlink function in Rust](https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html#tymethod.nlink) returns a 64-bit value.

This will increase the size of `__wasi_filestat_t`, however that seems negligible.

For comparison, the userspace `nlink_t` is already a `uint64_t`: https://github.com/rust-lang/libc/blob/master/src/wasi.rs#L34.

peterhuene:
This commit changes the type for `__wasi_linkcount_t` to 64-bits to support
large link counts.

Fixes #70.

### DIFF
--- a/phases/ephemeral/docs/wasi_ephemeral_preview.md
+++ b/phases/ephemeral/docs/wasi_ephemeral_preview.md
@@ -1799,7 +1799,7 @@ Members:
 
     The address and length of the buffer to be filled.
 
-### <a href="#linkcount" name="linkcount"></a>`__wasi_linkcount_t` (`uint32_t`)
+### <a href="#linkcount" name="linkcount"></a>`__wasi_linkcount_t` (`uint64_t`)
 
 Number of hard links to an inode.
 
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -451,7 +451,7 @@
 )
 
 ;; Number of hard links to an inode.
-(typename $linkcount_t u32)
+(typename $linkcount_t u64)
 
 ;; File attributes.
 (typename $filestat_t


peterhuene:
It looks like the default shell on Windows changed from cmd to psh?

alexcrichton:
Ah I think so yeah, you can fix it by inserting `shell: bash` just under to use bash on all platforms

peterhuene:
I opted for breaking the command into multiple lines rather than `&&`.  Any preference?

.github/workflows/main.yml
```diff
@@ -11,7 +11,9 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - name: Install Rust (rustup)
-      run: rustup update stable --no-self-update && rustup default stable
+      run: |
+        rustup update stable --no-self-update
+        rustup default stable
```
bjorn3: Is `set -e` used? Aka does one failing command cause the execution to stop?

alexcrichton:
I'd be fine either way, but I think @bjorn3 is right in that unfortunately `set -e` isn't the default. That may have changed though!

.github/workflows/main.yml
peterhuene: It appears that `sh` is the default shell for Linux and macOS, which is `-e`.  Setting the shell to `bash` will set `-eo pipefail`.

I don't think PowerShell has any such concept, but I think it behaves like -e does (that is, it will exit with failure on the first cmdlet that throws an exception).

However, I think it would probably be best to set `bash` as the shell so that we don't run into future incompatibilities.

PR #129, opened 2019-10-26
Fix the missing `flags` field in `subscription_clock_t`.

sunfishcode:
Add the `flags` field to `subscription_clock_t`, which was accidentally
omitted. This change reflects the definitions used in [existing practice],
so this patch updates the `wasi_unstable_preview0` and `wasi_unstable`
definitions as well.

[existing practice]: https://github.com/CraneStation/wasi-libc/blob/master/libc-bottom-half/headers/public/wasi/core.h#L421


peterhuene:
This should pass with a rebase.

PR #130, opened 2019-10-26
Document the special meaning of `dircookie_t` value 0.

sunfishcode:
Add a comment mentioning the special value of `dircookie_t`.


peterhuene:
This should pass with a rebase.

PR #131, opened 2019-10-28
Fix CI on Windows.

peterhuene:
Use `bash` as the default shell for installing Rust as `&&` is not a valid
command separator in PowerShell.


peterhuene:
Moving the CI fix out of #127 so that we can rebase the other PRs to get CI green.

sunfishcode:
Awesome, thanks!

Issue #132, opened 2019-11-02
Audit constants in the documentation


cjihrig:
Some of the constants in the documentation are out of order, based on what their values are. For example, the `__wasi_advice_t` values currently list `__WASI_ADVICE_DONTNEED` first, so I'd expect its value to be zero or possibly one. However, it's actually [four](https://github.com/CraneStation/wasi-libc/blob/ec3ee5e98541a89b9db016005fa0e9f0287fd755/libc-bottom-half/headers/public/wasi/core.h#L39). A [similar change](https://github.com/WebAssembly/WASI/commit/f606cfd7c1d20bf2954879ff09cfd463c1044093) also recently landed related to clock ID constants. It might be good to audit all of the constants for similar inconsistencies.

sunfishcode:
I believe #135 fixes this. If any further discrepancies arise, please report them!

pchickey:
Just to note here - we've informally discussed the possibility of moving the canonical definition of the docs into the witx file, and rendering them into markdown for human consumption. This may make it easier to maintain accurate docs.

syrusakbary:
> we've informally discussed the possibility of moving the canonical definition of the docs into the witx file, and rendering them into markdown for human consumption

That sounds good. As a comment it seems a bit odd that this things gets discussed in side-channels outside the official one.

Either way I'd like Wasmer be involved on this formal/informal discussions (since Wasmer is one of the biggest users of WASI).
Shall I setup some communication channel to make that easier?

PR #135, opened 2019-11-04
Sync WASI constants, types, and documentation with existing practice

sunfishcode:
Pursuant to https://github.com/WebAssembly/WASI/issues/132, I've now made a pass through all the types and constants, and have updated everything in this PR.


sunfishcode:
I should also mention, this modifies `old` and `unstable`, as it reflects existing practice.

PR #137, opened 2019-11-04
Mechanical witx naming convention changes

pchickey:
Witx files started out encoding the types directly from the wasi/core.h header, including the C rules around identifier capitalization and namespacing.

This PR drops those conventions that are specific to the C language. We do so uniformly in each witx file, meaning we change both the current unstable and the contents of `phases/old`, so that code generators that know about these new conventions can still support all witx files in this repo.

I performed the following changes, mechanically when possible (leaving the sed script in the commit message):
* Drop trailing `_t` from all typenames.
* Change all $SHOUTY_SNAKE_CASE identifiers to $snake_case.
* Remove repeated prefixes from flag and enum variants: e.g. drop right_ prefix on all $rights
    * Most cases of the above was removing the `thistypename_` prefix, with underscore to make it obvious. These were done by sed where possible.
    * Special cases: All `$errno` enum cases had the leading `e` dropped, all `$signal` enum cases had the leading `sig` dropped.

### DIFF
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -18,16 +18,16 @@
   (enum u32
     ;;; The clock measuring real time. Time value zero corresponds with
     ;;; 1970-01-01T00:00:00Z.
-    $clock_realtime
+    $realtime
     ;;; The store-wide monotonic clock, which is defined as a clock measuring
     ;;; real time, whose value cannot be adjusted and which cannot have negative
     ;;; clock jumps. The epoch of this clock is undefined. The absolute time
     ;;; value of this clock therefore has no meaning.
-    $clock_monotonic
+    $monotonic
     ;;; The CPU-time clock associated with the current process.
-    $clock_process_cputime_id
+    $process_cputime_id
     ;;; The CPU-time clock associated with the current thread.
-    $clock_thread_cputime_id
+    $thread_cputime_id
   )
 )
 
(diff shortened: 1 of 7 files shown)


phases/ephemeral/witx/typenames.witx
```diff
@@ -5,735 +5,735 @@
...
     ;; The file descriptor on which to wait for it to become ready for reading or writing.
-    (field $file_descriptor $fd_t)
+    (field $file_descriptor $fd)
   )
 )
 
-;; The contents of a $subscription_t.
-(typename $subscription_u
+;; The contents of a $subscription.
+(typename $u
```
sunfishcode: This ended up as a type named `$u` at the top-level, which would be good to avoid.

phases/ephemeral/witx/typenames.witx
```diff
@@ -742,10 +742,10 @@
 )
 
 ;; Information about a pre-opened capability.
-(typename $prestat_t
+(typename $prestat
   (struct
     ;; The type of the pre-opened capability.
-    (field $pr_type $preopentype_t)
+    (field $prype $preopentype)
```
sunfishcode: `$prype` looks like a `_t` rule applied in the middle of `$pr_type`.

phases/ephemeral/witx/typenames.witx
```diff
@@ -5,735 +5,735 @@
...
+    (field $d_next $dircookie)
     ;; The serial number of the file referred to by this directory entry.
-    (field $d_ino $inode_t)
+    (field $d_ino $inode)
     ;; The length of the name of the directory entry.
-    (field $d_namlen $dirnamlen_t)
+    (field $d_namlen $dirnamlen)
     ;; The type of the file referred to by this directory entry.
-    (field $d_type $filetype_t)
+    (field $dype $filetype)
```
sunfishcode: `$dype` looks like a `_t` rule applied in the middle of `$pr_type`.

pchickey:
Review feedback and merge conflicts are fixed

pchickey:
Rebased on #140. There were more conflicts than were worth resolving by hand, so I re-ran the sed commands and then re-applied the fixups.

PR #140, opened 2019-11-05
witx: Parse comments into documentation in AST

pchickey:
Proof of concept to show that https://github.com/alexcrichton/wat/pull/26/ is a viable approach


pchickey:
Todo:
* [ ] Only parse doc comments (`;;; line comments` or `(;; block comments ;;)`) into docs; ignore other comments.
* [ ] Trim whitespace
* [ ] Parse interior of doc comments as Markdown?

tools/witx/src/parser.rs
```diff
@@ -200,12 +281,14 @@ pub enum TopLevelSyntax<'a> {
 
 impl<'a> Parse<'a> for TopLevelSyntax<'a> {
     fn parse(parser: Parser<'a>) -> Result<Self> {
-        if parser.peek::<kw::r#use>() {
-            parser.parse::<kw::r#use>()?;
-            Ok(TopLevelSyntax::Use(parser.parse()?))
-        } else {
-            Ok(TopLevelSyntax::Decl(parser.parse()?))
-        }
+        parser.parens(|p| {
```
alexcrichton: It might be a bit easier if the `parser.parens` were moved directly inside `Documented` perhaps? Or maybe something like `Documented` and then like `DocumentedIdent` for enums/flags/etc

pchickey: The parens combinator was inside `Documented` but I had to move it out. Some items that need to be Documented are ordinary identifiers (enum & flag variants) not surrounded by parens.

pchickey:
I've got this PR to a point where I think it should be reviewed and merged:
* Doc comments are parsed at every node in the AST where they make sense to exist (imo). They are also output in the `Render` trait so that roundtrip testing still works. Unfortunately, whitespace doesn't roundtrip perfectly, and rather than come up with an elegant fix, I took the lazy way out and converted all the whitespace in our witx files so that it will round trip.
* A very rudimentary markdown output is implemented. This will need some polishing before we can replace the currently hand-written docs with automated docs. However, I think it is good enough to merge this code, in particular because the next point causes merge conflicts with essentially everything:
* The existing comments in all our witx files have been converted to doc-comments. The sed script that does so is included in the commit message. A few places that mechanical conversion didn't work are fixed up manually.

phases/ephemeral/witx/typenames.witx
```diff
@@ -1,767 +1,767 @@
-;; Type names used by low-level WASI interfaces.
+;;; Type names used by low-level WASI interfaces.
 ;;
```
alexcrichton: To confirm, I think the intention is that this is `;;;`, right? (perhaps just a sed script gone wrong?)

pchickey: ah, yeah, I missed that the very top comment on each file is not a doc. fixed.

pchickey:
Merge conflicts resolved. After we do a snapshot, I'll resolve the conflicts one more time and then merge.

pchickey:
Just rebased on #145 - will merge once that one is in

PR #141, opened 2019-11-05
Correct the return type of `fd_filestat_get`.

sunfishcode:
This fixes the old versions too, as this change reflects current
practice.


sunfishcode:
This is a fairly obvious fix, so I'm going to go ahead and merge.

PR #144, opened 2019-11-06
Fix old definition of `path_symlink`

alexcrichton:
This fixes the `*.witx` for the old definition of `path_symlink` to
ensure that the `old_path` argument comes first.


alexcrichton:
This was discovered when procedurally generating the Rust wasi bindings -- https://github.com/CraneStation/rust-wasi/pull/16

sunfishcode:
Good catch. Can you update the `unstable` and `ephemeral` witx files too?

alexcrichton:
Sure thing, done!

PR #145, opened 2019-11-07
witx: add support for parsing multiple top-level files

pchickey:
This PR adds support in witx to parse multiple top-level witx files. This is an api-breaking change - public functions that previously took an `impl AsRef<Path>` to a witx file now take an `&[impl AsRef<Path>]`

There is still only one global namespace for all type names and module names across all files. This means that if two top-level files define the same name, the second definition will cause an error, just like it would if both definitions are in the same file. However, if two top-level files depend each depend on a third file which defines a name, the name will only be defined once, and each reference to that name will be equal.

I'd like to fix the scoping issue at some point soon, but for now this unblocks #98.


pchickey:
Great idea - done :)

PR #146, opened 2019-11-08
Add a .gitignore and add Cargo.toml features

sunfishcode:
Add a basic .gitignore for working with Rust code, and add cargo features for witx's Cargo.toml.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+*.bk
+*.swp
+*.swo
+*.swx
+tags
+target
+Cargo.lock
+.*.rustfmt
+rusty-tags.*
+*~
+\#*\#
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -0,0 +1,11 @@
+*.bk
+*.swp
+*.swo
+*.swx
+tags
```
sbc100: I kind of feel like these wide gitignore rules for things like editor-specific backup names are probably better off in `$HOME/.gitignore` .. that way they effect all your projects and each user is in control of their own.

peterhuene: I think the point of having them in the repo's gitignore is that we would never want these files in the repo ever, regardless of how (or if) a contributor has configured their global gitignore. 

I personally think putting the common-enough ones in the repo's gitignore is a good way to keep unwanted files out; same reason as I would want `thumbs.db` and `.DS_Store` in every repo's gitignore as one never wants those files in a git repo.

sbc100: True, my only problem is that "common-enough" is not well defined and just because your editor uses `.swap` rather than `.swp` doesn't mean you should get to add your suffix to *every* git project you ever touch.    It seems to me that that is what your `$HOME/.gitignore` is for.

PR #147, opened 2019-11-08
Create snapshot 1

### LINKED ISSUE #138: Publishing a snapshot
We've had several PRs to the witx files, so it's time to start thinking about making an "unstable" snapshot, according to the [phases process](https://github.com/WebAssembly/WASI/blob/master/phases/README.md).

 - I've gotten feedback from a few folks that "unstable" is confusing. Should we rename "unstable" to "snapshot" while we have a chance here?
 - Are there any issues which are close to being resolved that we should wait for?

pchickey:
This PR creates the first snapshot of WASI. Since this is our first one since adopting phases in #88, there are a couple of changes, most significantly changing the name of "unstable" to "snapshot".

Running down the changes, commit-by-commit:

1. Move the contents of `phases/old/` to `phases/old/snapshot_0`. This makes it straightforward to archive the current snapshot next time around without changing any paths.
1. Delete contents of `phases/unstable`. For the special case of this first snapshotting, these are functionally identical to the contents of `phases/old/snapshot_0`, so they don't need to be archived.
1. Copy contents of `phases/ephemeral` into `phases/snapshot`
1. Rename files in `phases/snapshot` to substitute `ephemeral` for `snapshot`, and append snapshot number `1`.
1. Update the new phase naming convention in `phases/README.md`, and details the steps that should be sufficient to perform the next snapshot.
1. Update the witx crate tests according to new paths. All witx specs in tree still get validated.

Closes #138


phases/README.md
```diff
@@ -8,22 +8,25 @@
...
-   - Move all files supporting the APIs out of `ephemeral` into `unstable`.
-   - Rename the API modules from `wasi_ephemeral_` to `wasi_unstable_`
-     and append a version number.
+    1. `git mv` contents of `phases/snapshot/` to `phases/old/snapshot_{old_snapshot_number}`.
+    1. `cp -R` contents of `phases/ephemeral/` into `phases/snapshot/`.
+    1. Rename files copied into `phases/snapshot/` to substitute `ephemeral` for `snapshot` in file names. Append the new snapshot number to each name.
+    1. Update module names given in `.witx` files according to the previous step.
+    1. Update tests in `tools/witx/tests/wasi.rs` to point at new snapshot, and add a test
+       pointing at the just-archived snapshot under `old`.
+    1. Optionally, under `phases/old/snapshot_{old_snapshot_number}, add polyfills for suberseded APIs using the new APIs.
```
sbc100: One way we could add script to do this perhaps?

(I tent to prefer the numbered lists to actually increment the item number as it makes the md for readable/useful,  but I guess that is mostly bikeshed color preference).

pchickey: I worry that creating a script would be fragile because the places where you put the version number are subjective on what files are "top level" (e.g. not typenames), and #98 will make a bunch more top-level modules.

I'll update the markdown, doesnt matter to me either way

phases/README.md
```diff
@@ -8,22 +8,30 @@
...
+    2. `cp -R` contents of `phases/ephemeral/` into `phases/snapshot/`.
+    3. Rename files copied into `phases/snapshot/` to substitute `ephemeral`
+       for `snapshot` in file names. Append the new snapshot number to each
+         name.
+    4. Update module names given in `.witx` files according to the previous
+       step.
+    5. Update tests in `tools/witx/tests/wasi.rs` to point at new snapshot, and
+       add a test pointing at the just-archived snapshot under `old`.
+    6. Optionally, under `phases/old/snapshot_{old_snapshot_number}, add
+       polyfills for suberseded APIs using the new APIs.
```
sunfishcode: typo: suberseded

phases/README.md
```diff
@@ -8,22 +8,30 @@
...
+
 
   Pull Requests may also add additional tests, documentation, or
-  polyfills for existing `unstable` APIs.
+  polyfills for existing `snapshot` APIs.
 
 - [`old`](old): When APIs in `current` spec are replaced by new
```
sunfishcode: We can rename this `current` to `snapshot` too.

PR #152, opened 2019-11-19
witx: Depend on wast 3.0.4 to fix build failure with 3.0.1

joshtriplett:
witx depends on wast 3.0.1, but it uses the comment() method of the
Cursor type, which doesn't exist in version 3.0.1. This results in a
build failure if attempting to build witx against wast 3.0.1. That
method exists in the current 3.0.4, so update the wast dependency to
3.0.4.

### DIFF
--- a/tools/witx/Cargo.toml
+++ b/tools/witx/Cargo.toml
@@ -19,5 +19,5 @@ path = "src/main.rs"
 
 [dependencies]
 clap = "2"
-wast = "3.0.1"
+wast = "3.0.4"
 failure = "0.1"


pchickey:
Thanks for the fix. Im unclear on how we passed CI with this bug, will investigate.

pchickey:
Semver means cargo selects 3.0.4 at Cargo.lock creation time, with the latest crates.io. That explains why it works on my machine and in CI. If there is some way to detect the bound being too low automatically, let me know and I'll add it to CI.

alexcrichton:
This is a known deficiency in cargo and there's no stable solution at this time unfortunately

PR #175, opened 2019-12-16
Add a `usize` type to indicate types which will change size on wasm64.

sunfishcode:
This will allow code generators to know which types are sensitive to
wasm32 vs wasm64 without special-casing the `size` type.


sbc100:
With this translate to `ssize_t` in the C header?

sunfishcode:
Yes, this is `size_t` in C.

Theoretically there could be platforms where `size_t` differs from `uintptr_t`, but it seems quite unlikely that we'd do that with WASI.

PR #180, opened 2019-12-17
Document why there is no path_filestat_set_size

marmistrz:
As suggested by @sunfishcode in https://github.com/bytecodealliance/wasmtime/issues/273

### DIFF
--- a/phases/ephemeral/docs.md
+++ b/phases/ephemeral/docs.md
@@ -358,8 +358,13 @@ The right to invoke `path_rename` with the file descriptor as the target directo
 The right to invoke `path_filestat_get`.
 
 - <a href="#rights.path_filestat_set_size" name="rights.path_filestat_set_size"></a> `path_filestat_set_size`
-The right to change a file's size (there is no `path_filestat_set_size`).
+The right to change a file's size.
 If `path_open` is set, includes the right to invoke `path_open` with [`oflags::trunc`](#oflags.trunc).
+Note: there is no function named `path_filestat_set_size`. This follows POSIX design,
+which only has `ftruncate` and does not provide `ftruncateat`.
+While such function would be desirable from the API design perspective, there are virtually
+no use cases for it since no code written for POSIX systems would use it.
+Moreover, implementing it would require multiple syscalls, leading to inferior performance.
 
 - <a href="#rights.path_filestat_set_times" name="rights.path_filestat_set_times"></a> `path_filestat_set_times`
 The right to invoke `path_filestat_set_times`.
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -249,8 +249,8 @@
     $path_filestat_get
     ;;; The right to change a file's size.
     ;;; If `path_open` is set, includes the right to invoke `path_open` with `oflags::trunc`.
-    ;;; Note: there is no `path_filestat_set_size`. This follows POSIX design, which only has
-    ;;; `ftruncate` and does not provide `ftruncateat`.
+    ;;; Note: there is no function named `path_filestat_set_size`. This follows POSIX design,
+    ;;; which only has `ftruncate` and does not provide `ftruncateat`.
     $path_filestat_set_size
     ;;; The right to invoke `path_filestat_set_times`.
     $path_filestat_set_times
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -247,8 +247,10 @@
     $path_rename_target
     ;;; The right to invoke `path_filestat_get`.
     $path_filestat_get
-    ;;; The right to change a file's size (there is no `path_filestat_set_size`).
+    ;;; The right to change a file's size.
     ;;; If `path_open` is set, includes the right to invoke `path_open` with `oflags::trunc`.
+    ;;; Note: there is no `path_filestat_set_size`. This follows POSIX design, which only has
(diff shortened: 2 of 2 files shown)


phases/ephemeral/witx/typenames.witx
```diff
@@ -248,8 +248,10 @@
     $path_rename_target
     ;;; The right to invoke `path_filestat_get`.
     $path_filestat_get
-    ;;; The right to change a file's size (there is no `path_filestat_set_size`).
+    ;;; The right to change a file's size.
     ;;; If `path_open` is set, includes the right to invoke `path_open` with `oflags::trunc`.
+    ;;; Note: there is no `path_filestat_set_size`. This follows POSIX design, which only has
+    ;;; `ftruncate` and does not provide `ftruncateat`.
```
sbc100: I found this confusing to read because there clearly is a thing `path_filestat_set_size` called on the line below.  I guess you are referring to the fact that there is no addition separate function with this name?

sbc100:
The fact that POSIX lacks a `truncateat` function seems like it was probably more of an oversight than a deliberate decision.    Begin able to set a files size directly without opening seems useful to me.

Also, IIUC we are explicitly not following POSIX's design so the comment seems a little strange.

phases/ephemeral/witx/typenames.witx
marmistrz: Yes, the intended meaning is that there is no function named `__wasi_path_filestat_set_size`.

marmistrz:
@sbc100 
Lack of `ftruncateat` means that, as far as I know, there is no way to efficiently implement `path_filestat_set_size` on any major platform.

On Unix the only possible implementation seems to be `path_open` + `fd_filestat_set_size` corresponding to `open` + `ftruncate`.
On Windows there is `SetEndOfFile` which similarly expects a file handle, so again we end up doing `path_open` + `fd_filestat_set_size`.

Exposing `path_filestat_set_size` could give an illusion that this call is cheaper than `path_open` + `fd_filestat_set_size`.

phases/ephemeral/witx/typenames.witx
sunfishcode: Could you change the text to add the word "function" here, so that it's clear you're talking about a function by that name (which indeed doesn't exist) rather than a right (which does)?

marmistrz: Done!

phases/ephemeral/witx/typenames.witx
```diff
@@ -248,8 +248,10 @@
     $path_rename_target
     ;;; The right to invoke `path_filestat_get`.
     $path_filestat_get
-    ;;; The right to change a file's size (there is no `path_filestat_set_size`).
+    ;;; The right to change a file's size.
     ;;; If `path_open` is set, includes the right to invoke `path_open` with `oflags::trunc`.
+    ;;; Note: there is no function named `path_filestat_set_size`. This follows POSIX design,
+    ;;; which only has `ftruncate` and does not provide `ftruncateat`.
```
sbc100: I see.  This makes more sense to me now.

Perhaps you could rephrase.  This comment makes it sound like we don't have this function because we are following POSIX, which is not that case.  How about something like this:

```
Note: there is no function named `path_filestat_set_size` because on many existing
systems there is no way to efficiently implement this.  For example, POSIX has no
`ftruncateat` function.
```

marmistrz: Sorry for the delay. I rephrased it in a slightly different way, is it okay for you now?

marmistrz:
How should I regenerate `/home/runner/work/WASI/WASI/tools/witx/../../phases/ephemeral/witx/../docs.md`? This is why the CI fails.

phases/ephemeral/docs.md
```diff
@@ -337,8 +337,12 @@ The right to invoke `path_rename` with the file descriptor as the target directo
 The right to invoke `path_filestat_get`.
 
 - <a href="#rights.path_filestat_set_size" name="rights.path_filestat_set_size"></a> `path_filestat_set_size`
-The right to change a file's size (there is no `path_filestat_set_size`).
+The right to change a file's size.
 If `path_open` is set, includes the right to invoke `path_open` with [`oflags::trunc`](#oflags.trunc).
+Note: there is no function named `path_filestat_set_size`. This follows POSIX design,
+which only has `ftruncate` and does not provide `ftruncateat`. While such function
+would definitely be useful, it would be difficult to implement `path_filestat_set_size`
+efficiently on most existing systems.
```
sunfishcode: Would it be difficult to implement efficiently? It seems we could implement this by just opening the file and doing `ftruncate`, which would be reasonable. My understanding is that the main reason we don't have `path_filestat_set_size` is that we don't currently have any use cases for it, since no code written for POSIX systems would use it.

marmistrz: Here _to implement efficiently_ is intended to mean _to implement using one syscall only_, as we consider the cost of other calls negligible.

How about 
> While such function would be desirable from the API design perspective, there are virtually no use cases for it since no code written for POSIX systems would use it. Moreover, implementing it would require multiple syscalls, leading to inferior performance.

marmistrz: @sunfishcode bump

sunfishcode: Thanks for the bump! That wording sounds good to me.

marmistrz: I pushed the new wording.

PR #189, opened 2020-01-01
Add FIFO filetype to match POSIX

vapier:
Every other filetype defined by POSIX for stat is in here,
so add FIFO to round things out.


sbc100:
Seems reasonable to me.  @sunfishcode ?

PR #192, opened 2020-01-06
Add a WASI design principles document.

sunfishcode:
As part of a larger effort to answer questions about what capability-based design means, whether WASI is intending to be POSIX-oriented, and other things, here's a draft of a design principles document.

This also includes a section on virtualization, which takes some ideas from #69, but modifies it to focus on using interface types and the linking proposal to implement virtualization.

Feedback is welcome!

Edit: [Click here to read as rendered html](https://github.com/WebAssembly/WASI/blob/7d6a36e4f52c3965b8e8732061e1b9642e2bb276/docs/DesignPrinciples.md)


docs/WASI-design-principles.md
```diff
@@ -0,0 +1,93 @@
...
+
+If all instances shared a single c-list, that would undermine isolation, because any instance could forge an index to any other instance's resources. Fortunately, there can be multiple c-lists, and instances can be given access to some c-lists and not others, or even have instances which don't use any c-lists at all, so it's still possible to establish isolation between instances.
+
+Witx-specified APIs use a special `handle` keyword to mark parameters and return values which are handles. In the short term, these are lowered to integer indices, with an implied table, so that the APIs can be easily used from C and similar languages today. Once [interface types](https://github.com/WebAssembly/interface-types) and [type imports](http://github.com/WebAssembly/type-imports-proposal) are ready, we expect to make use of them to provide APIs which can be used either from languages using references or from languages using integer indices, with tables being used and managed automatically.
```
jacobrosenthal: noticed a broken link, should be https://github.com/WebAssembly/proposal-type-imports

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,93 @@
...
+
+That said, where other considerations don't interfere, WASI should use existing Web standards and work with Web standardization efforts rather than gratuitously inventing its own versions of them and/or duplicating efforts.
+
+When using Web standards, WASI APIs should be careful to avoid depending on JavaScript in the engine in APIs where it isn't essential.
+
+## Use WebAssembly standards and proposals
+
+WASI should align with and build on WebAssembly standards and proposals where applicable.
+
+For example, WASI seeks to align with and build on [interface types](https://github.com/WebAssembly/interface-types), [multiple return values](https://github.com/WebAssembly/multi-value/), [reference types](https://github.com/WebAssembly/reference-types), [type imports](http://github.com/WebAssembly/type-imports-proposal), and more. As of this writing, some of these are early-stage proposals, so we're not actually depending on them yet, however we are carefully aligning with them so that we'll be ready when they are.
```
jacobrosenthal: noticed a broken link, should be https://github.com/WebAssembly/proposal-type-imports

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,93 @@
...
+
+ - Shared filesystem views - One of the unique capabilities WebAssembly brings to the table is the possibility of shared-nothing linking between applications and libraries. Shared-nothing means that all communication is via explicit calls, and the libraries don't share an address space or any other implicit shared state. But if we run both sides within the same filesystem view, that would give them a large body of shared state. Union mounts, mandatory access control systems, user namespaces, and other techniques can help, but often require complex configuration, heavy-weight boundaries, and sometimes even admin privileges to set up.
+
+This has wide-ranging implications. Much of POSIX is oriented around passing around strings, whether through command-line arguments, environment variables, or paths embedded in files, with the assumption that there's a shared filesystem view between components. As we said above, we're de-emphasizing strings, which dovetails with de-emphasizing shared filesystem views.
```
tlively: It might be useful to explain what pattern WASI will use to share files and data in the absence of a shared filesystem view. Will applications be expected to explicitly pass file access capabilities for each relevant file to each relevant dependency?

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,93 @@
...
+
+Witx-specified APIs use a special `handle` keyword to mark parameters and return values which are handles. In the short term, these are lowered to integer indices, with an implied table, so that the APIs can be easily used from C and similar languages today. Once [interface types](https://github.com/WebAssembly/interface-types) and [type imports](http://github.com/WebAssembly/type-imports-proposal) are ready, we expect to make use of them to provide APIs which can be used either from languages using references or from languages using integer indices, with tables being used and managed automatically.
+
+## WASI's Scope
+
+WASI started out with a very POSIX-like API, wowever, WASI will grow to include many APIs that are outside of the scope of POSIX. WASI is a forum for cooperatively designing APIs, along with a W3C CG Subgroup for eventually standardizing them.
```
programmerjake: wowever should be however

sunfishcode: Fixed, thanks!

docs/WASI-design-principles.md
sunfishcode: Fixed, thanks!

docs/WASI-design-principles.md
sunfishcode: Good idea. I've now added some extra text to illustrate this.

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,95 @@
...
+
+If all instances shared a single c-list, that would undermine isolation, because any instance could forge an index to any other instance's resources. Fortunately, there can be multiple c-lists, and instances can be given access to some c-lists and not others, or even have instances which don't use any c-lists at all, so it's still possible to establish isolation between instances.
+
+Witx-specified APIs use a special `handle` keyword to mark parameters and return values which are handles. In the short term, these are lowered to integer indices, with an implied table, so that the APIs can be easily used from C and similar languages today. Once [interface types](https://github.com/WebAssembly/interface-types) and [type imports](https://github.com/WebAssembly/proposal-type-imports) are ready, we expect to make use of them to provide APIs which can be used either from languages using references or from languages using integer indices, with tables being used and managed automatically.
+
```
kripken: Thoughts on this proposal https://github.com/WebAssembly/WASI/issues/109#issuecomment-542921814 - would that fit in this model?

kripken: Specifically, I think it would be good to address how the standard POSIX etc. model of "get me a handle for this string path" would work - who would be responsible for that, would it be in the VM or in wasm or in a polyfill (as in that proposal) or somewhere else.

sunfishcode: I added some text [here](https://github.com/WebAssembly/WASI/pull/192/commits/baaa84fa878ad86eac2fa661662dd3716a7e8a30) which gives an overview of how this would look. Does that answer your question?

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,95 @@
...
+For example, WASI may include high-level network APIs, such as APIs for HTTP. This is outside the scope of POSIX, and while some WebAssembly engines are very interested in implementing it natively, others will find it too complex and high-level. But one of the great things about WebAssembly is that there's no syscall instruction, so "syscalls" in WebAssembly are just calls to imported functions, which could be native functions provided by the runtime, or could be other WebAssembly modules. We expect to leverage this capability to provide polyfill implementations of things like high-level network APIs on top of low-level APIs, such as a raw socket API, so that engines which wish to keep things simple and just implement the low-level socket APIs can do so.
+
+WASI also aims to include domain-specific APIs, such as database, blockchain, or specialized APIs for embedded systems. Another key building block for WASI is optional imports, which give applications the ability to dynamically test for the availability of APIs.
```
eddyp: Would you want to be able to ask "Tell me all the available APIs"? I'd say that would be a security risk and a possible resource hog, so I'd go for "No", and, instead, explicitly asking for "Do you have API X?" is more secure and seems to make sense from and application PoV.

docs/WASI-design-principles.md
kripken: Not fully: it has words like "de-emphasizing" and "typically", so I'm not sure how to interpret it. Concretely, again, I'm asking if this document would accept @tlively's compromise proposal, or not (or "unsure").

docs/WASI-design-principles.md
sunfishcode: Yes, "do you have API X" is exactly what optional imports do.

docs/WASI-design-principles.md
tlively: Specifically, I think the ambiguity comes from the sentence `When needed, compatibility with POSIX-style APIs is then provided as a thin layer on top implementing a simple name-to-handle mapping.` This sentence does not indicate which party should be responsible for providing this layer. There would be a benefit to code size and portability if WASI environments had to provide these compatibility layers as standardized imports in much the same way that normal WASI functions are standardized imports. The difference would be that the compatibility layers would be implemented in user space rather than by the engine to preserve OCAP.

sunfishcode: Are you talking about [this](https://github.com/WebAssembly/WASI/issues/109#issuecomment-542921814)? I'll respond in that issue.

tlively: Yes, that's what we're thinking of. Follow-up discussion there sounds good.

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,95 @@
+# WASI Design Principles
+
```
sbc100: Maybe just call this file `design-principles.md`?   It should be obvious from the repo it lives that its about WASI.  Otherwise every file in this directory will end being called WASI-something, which I think we want to above since its redundant.

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,95 @@
...
+## Capability-based security
+
+WASI is built using capability-based security principles. Access to external resources is always represented by *handles*, which are special values that are *unforgeable*, meaning there's no way to coerce an arbitrary integer or other type of value into a handle. WASI is also aiming to have no *ambient authorities*, meaning that there should be no way to request a handle purely by providing a string or other user-controlled identifier providing the name of a resource. With these two properties, the only ways to obtain access to resources are to be explicitly given handles, or to perform operations on handles which return new handles.
```
sbc100: Would you mind wrapping these long paragraphs at 80 chars?

sbc100: `special values that are *unforgeable*`.. this isn't true today, which I think warrants mentioning here.

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,95 @@
...
+
+The simplest representation of handles are values of [reference type](https://github.com/WebAssembly/reference-types). References in wasm are inherently unforgeable, so they can represent handles directly.
+
+Some programming languages operate primarily within linear memory, such as C, C++, and Rust, and there currently is no easy way for these languages to use references in normal code. And even if it does become possible, it's likely that source code will still require annotations to fully opt into references, so it won't always be feasible to use. For these languages, references are stored in a [table](https://webassembly.github.io/spec/core/bikeshed/index.html#tables%E2%91%A0) called a *c-list*. Integer indices into the table then identify resources, which can be easily passed around or stored in memory. In some contexts, these indices are called *file descriptors* since they're similar to what POSIX uses that term for. There are even some tools, such as wasm-bindgen, which make this fairly easy. (Internally, tools and engines don't always use actual WebAssembly tables to do this, however those are implementation details. Conceptually, they work as if they had tables.)
```
sbc100: Again, we should be clear up front that this doesn't actually describe the current implemenation... I see now that you do this below which maybe is fine, but it means that people who just glance at this section might get the wrong impression.

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,95 @@
...
+
+For example, a typical POSIX-style API might include a function that accepts a file name to open. That requires the implementation to have a filesystem view, and to have appropriate permissions within that filesystem view. WASI APIs typically prefer to instead have a function which accepts a handle for an already-open file. That way, the implementation doesn't need a filesystem view, or permissions within the filesystem. It doesn't even need to care whether there even is a filesystem. When needed, compatibility with POSIX-style APIs is then provided as a thin layer on top implementing a simple name-to-handle mapping.
+
+We recognize that this approach has tradeoffs. It often does take more work to design and implement the compatibility layers needed to support existing applications than if we just always made WASI always expose POSIX-style APIs directly. And, even when we do have compatibility mechanisms, they aren't always the most locally optimal ones. The compatibility layer overhead is usually quite modest, but it is present.
```
sbc100: Double use of "always" in the second sentence seems odd.  Perhaps rephrase that sentence?    Do you need the "does" in "It often does take more work"?

docs/WASI-design-principles.md
```diff
@@ -0,0 +1,95 @@
...
+
+However, shared-nothing linking makes it easier to use libraries in more circumstances, because you don't have to have the trust implied by a shared filesystem view, or the complexity of configuring filesystem rules for each library. With a better story for libraries and tools to work together in cooperation with the sandbox, we can build a stronger ecosystem which makes up for the downsides in the long run.
```
sbc100: "libraries" -> "programs" .. isn't shared nothing linkage really for for programs?    If you build a library with shared nothing linkage isn't it really program at that point?

sbc100: "libraries and tools" -> "programs"

docs/WASI-design-principles.md
sunfishcode: Done.

docs/WASI-design-principles.md
sunfishcode: Done.

sunfishcode: It's sort of true, in that in C/etc. the handles are conceptually still there, they're just stored in tables for you. You can't forge new handles that aren't already in your table.

This is discussed more a few paragraphs later; does that clarify things?

docs/WASI-design-principles.md
sunfishcode: I'm not clear on what you mean. The parenthetical "Internally, ..." does describe what effectively happens today -- there is no actual WebAssembly table, but things act as if there were.

docs/WASI-design-principles.md
sunfishcode: I cleaned up the "always". The "does" is for emphasis. This is is responding to certain people who publicly criticizing WASI over this; the intent is to acknowledge that there is a tradeoff being made here, and to argue that the tradeoff is worthwhile.

docs/WASI-design-principles.md
sbc100: Yes, I agree that it does become more clear with the followup paragraphs.   I just found it slightly misleading to open with "special values that are *unforgeable*" since the current state is of WASI is that (as far as I understand) FDs are exactly as un-forgable is POSIX FDs.    You can conjure them out of thin air and use them (as long as your process you are in already has such as FD open).   For this reason I think the unforgability part might be better of explicitly in the a forward looking section of this document.

docs/WASI-design-principles.md
sbc100: Agreed.

docs/WASI-design-principles.md
sunfishcode: On one hand, it's a good point. Shared-nothing linking does make "libraries" look like "programs" in a sense. On the other, the core concept of a "library" is code reuse, and that's specifically what I'm looking to highlight here. If you want to reuse code, including code written by other people, using shared-nothing linking to link in that code in -- as a library -- means you don't have to worry about the implicit effects it might have on your program state.

sbc100: What is the difference between a library and tool in shared nothing linkage?

In order to make library useful in the mode wouldn't you basically need to turn it into a reactor, which is then a tool isn't it?   I know we haven't officially adopted the reactor terminology, but what would the difference be between a reactor and library that is been exposed/wrapped via SNL?

I worry that people might think they can just take a set of C libraries and build them as SNL and expect to have them work together unmodified.

Having said that I do get why library here can make sense, so I won't object to it.

sunfishcode: The terminology is very much in flux, but as I see it, the key thing is instance lifetimes: A command has a lifetime scoped to a call frame. A reactor lives until something (possibly itself) terminates it. And a library is something that inherits a lifetime from the context it's used in.

It's a good point that shared-nothing linking isn't going to magically fall out just by recompiling existing C code. I'll add some more words to clarify that.

docs/WASI-design-principles.md
sunfishcode: We can still do shared-nothing linking though, even with libraries written in C, even when they use integer indices internally, because individual libraries can have their own tables. So it's true that C code could forge any index it wants, but it's still limited to accessing those things which have been placed in its table.

sunfishcode: I've now rewritten the paragraph that explains this, so that it now starts with "Integer indices are themselves forgeable...", which hopefully clarifies this.

docs/DesignPrinciples.md
```diff
@@ -0,0 +1,291 @@
...
+have a filesystem view, and to have appropriate permissions within
+that filesystem view. WASI APIs typically prefer to instead have a
+function which accepts a handle for an already-open file. That way, the
+implementation doesn't need a filesystem view, or permissions within
+the filesystem. It doesn't even need to care whether there even is a
+filesystem. When needed, compatibility with POSIX-style APIs is then
+provided as a thin layer on top implementing a simple name-to-handle
+mapping.
+
+We recognize that this approach has tradeoffs. It often does take more
```
frzifus:
```suggestion
We recognize that this approach has trade-offs. It often does take more
```
:octopus:

sunfishcode:
Thanks to everyone who reviewed this! I've addressed all the feedback, and there were no major objections, so I'll proceed to merge this. Feedback is still welcome; please file issues to ask questions or make comments, or even feel free to submit PRs to suggest changes.

PR #193, opened 2020-01-08
Make poll return errno::inval if there are no subscriptions.

sunfishcode:
`poll` with no subscriptions is effectively an infinite hang. Since wasm
has no signals, use cases which involve waking up a libc `pause` call, which
is implemented in terms of a poll with no subscriptions, aren't applicable.
So all `pause` can do is suspend the program until the host environment
terminates it.

One situation where that's useful is in debugging, as it's sometimes useful
to be able to suspend a process in place so that it can be inspected by a
debugger. However, that doesn't require a fully infinite suspend; a suspend
with a long timeout is adequate.

Making the no-subscriptions case an error can help catch cases where
applications unintentionally enter infinite loops.

### DIFF
--- a/phases/ephemeral/witx/wasi_ephemeral_poll.witx
+++ b/phases/ephemeral/witx/wasi_ephemeral_poll.witx
@@ -12,6 +12,8 @@
   (import "memory" (memory))
 
   ;;; Concurrently poll for the occurrence of a set of events.
+  ;;;
+  ;;; If `nsubscriptions` is 0, returns `errno::inval`.
   (@interface func (export "oneoff")
     ;;; The events to which to subscribe.
     (param $in (@witx const_pointer $subscription))


marmistrz:
I think that `ENOTSUP` would be more fortunate. On Linux this is a valid value for `poll(2)`:
> EINVAL (ppoll()) The timeout value expressed in *ip is invalid (negative).

But it's not supported by WASI.

sunfishcode:
That's a good point.

From a pure WASI perspective, this issue feels more like an `EINVAL` situation; then we'd just say that calling the `poll` syscall with no subscriptions is a bug, which is what `EINVAL` is for. But from a POSIX and Linux perspective the `poll` libc function does support `poll` with 0 subscriptions, so maybe we could have WASI libc return `ENOTSUP` in that case. Does that make sense?

sunfishcode:
We discussed this in the video call today, and there were no objections, so I'm calling it a consensus.

PR #197, opened 2020-01-10
Remove the cputime clocks.

sunfishcode:
Remove $process_cputime_id and $thread_cputime_id.

Some implementations will not have one wasm nanoprocess per OS process,
or one wasm thread per OS thread, making it difficult to implement
$process_cputime_id meaningfully. Also, the concept of "cputime"
available in most host environments isn't easy to directly connect to
what we might expect "cputime" to mean for wasm code.

These correspond to `CLOCK_PROCESS_CPUTIME_ID` and `CLOCK_THREAD_CPUTIME_ID`
in POSIX, which are optional:

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/time.h.html

### DIFF
--- a/phases/ephemeral/docs.md
+++ b/phases/ephemeral/docs.md
@@ -21,12 +21,6 @@ real time, whose value cannot be adjusted and which cannot have negative
 clock jumps. The epoch of this clock is undefined. The absolute time
 value of this clock therefore has no meaning.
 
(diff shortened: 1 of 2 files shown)


pchickey:
Please merge with `master` before merging so that this PR picks up CI checking for docs generation. You'll have to regenerate docs as well (`cd tools/witx && cargo run -- repo-docs`).

sunfishcode:
Merged with master and docs auto-generated.

PR #199, opened 2020-01-10
Rename `creat` to `create`.

sunfishcode:
Ken Thompson was once asked what he would do differently if he were
redesigning the UNIX system. His reply: "I'd spell creat with an e."

 - https://en.wikiquote.org/wiki/Ken_Thompson#Quotes

### DIFF
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -228,7 +228,7 @@
     $fd_allocate
     ;;; The right to invoke `path_create_directory`.
     $path_create_directory
-    ;;; If `path_open` is set, the right to invoke `path_open` with `oflags::creat`.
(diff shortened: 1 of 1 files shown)


sunfishcode:
I'm ok either way there. We do have a lot of other abbreviations, like `inval`, `isdir`, and so on, so if you want, there are a lot more we could consider here. I think `creat` just feels especially gratuitous because it's only saving one letter. And Ken Thompson would seem to agree ;-).

joshtriplett:
:+1:

The "creat" abbreviation does need to exist in libc and similar libraries, to avoid gratuitously breaking compatibility. However, we don't need that abbreviation in the native syscall layer.

sunfishcode:
I think another factor here is something like an uncanny valley effect -- `creat` just looks like a typo rather than an abbreviation.

In any case, we seem to have consensus for renaming `creat` at least, so let's do it!

PR #200, opened 2020-01-10
Rename $subscription_fd_readwrite's $fd member to $fd.

sunfishcode:
This makes it more consistent with naming through the rest of the witx
specs.

### DIFF
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -613,7 +613,7 @@
 (typename $subscription_fd_readwrite
   (struct
     ;;; The file descriptor on which to wait for it to become ready for reading or writing.
-    (field $file_descriptor $fd)
+    (field $fd $fd)
   )
 )
 


kubkon:
Oh oops, this one is definitely co-dependent on #203. I'll change #203 after we merge this one :+1:

pchickey:
Sorry, fat finger closed it by accident

kubkon:
@sunfishcode Since we've landed #203, would you mind changing lines [typenames.witx#L490] and [typenames.witx#L493] so that the links there now read `subscription_fd_readwrite::fd` instead of `subscription_fd_readwrite::file_descriptor`?

[typenames.witx#L490]: https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/witx/typenames.witx#L490
[typenames.witx#L493]: https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/witx/typenames.witx#L493

sunfishcode:
This patch only applies to the ephemeral directory; the field is still named `file_descriptor` in the snapshot directory.

kubkon:
> This patch only applies to the ephemeral directory; the field is still named `file_descriptor` in the snapshot directory.

Ahh, gotcha! I missed that one, sorry!

PR #203, opened 2020-01-15
Fix some links in the docs

kubkon:
This commit fixes some links in different places in the docs
sections in `snapshot1` `*.witx` files so that they are consistent
with type naming throughout the document. This will aid correct
Markdown link generation by the `witx` tool directly from `*.witx`
files.

Note that thanks to this fix, the links are now correctly generated by `witx`
tool with the new docs generator introduced in #194.


kubkon:
cc @sunfishcode

PR #205, opened 2020-01-16
Replace clap with structopt

kubkon:
IMHO, `structopt` generally makes things somewhat cleaner and is worth
moving `witx` to `structopt` from `clap` for added readability and
extendability of the tool in the future.

### DIFF
--- a/tools/witx/Cargo.toml
+++ b/tools/witx/Cargo.toml
@@ -18,7 +18,7 @@ name = "witx"
 path = "src/main.rs"
 
 [dependencies]
-clap = "2"
+structopt = "0.3"
 wast = "3.0.4"
 thiserror = "1.0"
 log = "0.4"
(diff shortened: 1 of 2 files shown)


kubkon:
cc @pchickey could you have a quick glance and check that I've ported all subcommands and args correctly please? :-)

kubkon:
@sunfishcode @pchickey I guess we can go ahead and merge this one in ourselves right? Sorry for the potentially redundant question, but I’d rather tread carefully here :-)

sunfishcode:
For tooling-only changes like this, regular approval from other folks who work in the area, as you have here, is fine. This PR adds a new dependency, but structopt is entirely reasonable. So feel free to merge!

kubkon:
> For tooling-only changes like this, regular approval from other folks who work in the area, as you have here, is fine. This PR adds a new dependency, but structopt is entirely reasonable. So feel free to merge!

Awesome, thanks for clarifying! :-)

PR #208, opened 2020-01-21
Fix typo in Charter.md

denisvasilik:


kubkon:
BTW @pchickey @sunfishcode, it seems that a change in the docs creeped in in the meantime hence why failing CI :-D

pchickey:
Fixed in #210, re-running CI now.

Issue #209, opened 2020-01-21
Function error code acces or access?


denisvasilik:
I just wondered whether the function error code should be named **access** rather than **acces**?

kubkon:
Yeah, that's a remnant from POSIX (there, it's called `EACCES`). Personally, much like in case of `oflags::creat` (see #199 for reference) we don't really save much in terms of the encoding, so I'd be keen to vote in favour of having `errno::access`. :-)

denisvasilik:
I see `oflags:creat` is another one. May be it makes sense to generally discuss 
naming conventions for WASI and put the result into the design principles document.
Personally, I think this is where WASI should distinguish itself from POSIX a bit. 
For this issue, I'm just going to create a pull request. Let's see what happens :)

sunfishcode:
The PR to rename `acces` to `access` is now merged! And `creat` is already renamed to `create` on the ephemeral branch.

If people are contemplating broader renaming efforts, please file new issues so that we can discuss them before doing the actual work. So far, renaming has been limited to roughly "abbreviations which look wrong rather than looking like abbreviations", so if we're going to expand the scope, we should talk about it first :-).

PR #211, opened 2020-01-22
Rename acces to access #209

denisvasilik:
This pull request is part of issue #209.


sunfishcode:
This makes sense to me. There's evidence that people commonly spell this one `access` despite the POSIX spelling:

https://lore.kernel.org/patchwork/patch/1003891/
https://lore.kernel.org/patchwork/patch/1079508/
https://lkml.org/lkml/2019/5/27/185

Of course, WASI libc will need to continue to expose this as `EACCES`, because that is the POSIX name. So the primary effect of this name change will be to make the documentation more clear, but that seems worth it to me here.

sunfishcode:
Thanks!

PR #212, opened 2020-01-26
change set_permissions to permissions_set

cjihrig:
It's not clear which is correct, but `pemissions_set` seems to be used more frequently.


phases/ephemeral/witx/typenames.witx
```diff
@@ -253,15 +253,15 @@
     $path_filestat_set_size
     ;;; The right to invoke `path_filestat_set_times`.
     $path_filestat_set_times
-    ;;; The right to invoke `path_filestat_set_permissions`.
+    ;;; The right to invoke `path_filestat_permissions_set`.
```
cjihrig: Actually, it looks like `filestat` shouldn't be part of these names either.

PR #213, opened 2020-01-26
update fd_permissions_set() signature

cjihrig:
It appears that the `permissions` argument to `permissions_set()` was labeled as an output instead of an input.

### DIFF
--- a/phases/ephemeral/docs.md
+++ b/phases/ephemeral/docs.md
@@ -308,7 +308,7 @@ The right to invoke `fd_allocate`.
 The right to invoke `path_create_directory`.
 
 - <a href="#rights.path_create_file" name="rights.path_create_file"></a> `path_create_file`
-If `path_open` is set, the right to invoke `path_open` with [`oflags::creat`](#oflags.creat).
+If `path_open` is set, the right to invoke `path_open` with [`oflags::create`](#oflags.create).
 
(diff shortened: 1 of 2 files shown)


kubkon:
Oh, and BTW, in order to make the CI green whenever introducing a change to the `*.witx` files in a PR, you should regenerate the docs and include them in the PR. You can do this by running

```
$ pushd tools/witx
$ cargo run -- repo-docs
$ popd
```

This will regenerate the docs for all phases which you can then commit in the PR :-)

cjihrig:
Thanks @kubkon. I noticed the link checking was failing, but also noticed that the witx and generated docs hadn't been kept in sync, so I wasn't sure if that was something I was supposed to do.

kubkon:
> Thanks @kubkon. I noticed the link checking was failing, but also noticed that the witx and generated docs hadn't been kept in sync, so I wasn't sure if that was something I was supposed to do.

Yep, we got behind on keeping our docs up-to-date with the state of the `*.witx` files somewhat but that is now fixed in #217. However, even if the docs were kept up-to-date, since you have modified the `*.witx` here by replacing `result` with `param` in one of `permissions_set`'s args, the CI would still be unhappy until you regenerated the docs with your change included :-)

cjihrig:
Ping. I pushed a commit to regenerate the docs.

PR #216, opened 2020-01-30
Reorder results of fd_permissions_set in ephemeral

ueno:
I tried to feed the current ephemeral witx files to the `generate-raw` tool and noticed a minor inconsistency. Unlike other interface functions, `permissions_set` in `wasi_ephemeral_fd.witx` returns `error` as the second result, which is not supported by the tool:
https://github.com/bytecodealliance/wasi/blob/master/crates/generate-raw/src/lib.rs#L280

Might it make sense to change the order to be consistent with the other part of the proposal?


cjihrig:
I think the issue is that this is supposed to be an input parameter. See https://github.com/WebAssembly/WASI/pull/213

ueno:
Ah, indeed. Thank you!

PR #217, opened 2020-01-30
Regenerate docs for wasi_ephemeral

kubkon:
This commit regenerates docs for the ephemeral phase since there
were a couple of changes to the underlying `*.witx` which started
tripping the CI. This commit fixes that.

### DIFF
--- a/phases/ephemeral/docs.md
+++ b/phases/ephemeral/docs.md
@@ -308,7 +308,7 @@ The right to invoke `fd_allocate`.
 The right to invoke `path_create_directory`.
 
 - <a href="#rights.path_create_file" name="rights.path_create_file"></a> `path_create_file`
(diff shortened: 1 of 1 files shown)


pchickey:
Looks like those are from PRs which branched off before #204 merged. There's only one of those left open that will break CI when merged (#197), I've left a comment there to please rebase or merge the latest master in before merging the PR.

sunfishcode:
I updated the docs in #197 before merging it, and it looks like the docs were already auto-generated for the other changes here, so everything is now up to date!

kubkon:
> I updated the docs in #197 before merging it, and it looks like the docs were already auto-generated for the other changes here, so everything is now up to date!

Makes sense, thanks!

Issue #221, opened 2020-02-06
Trivial change in the description: Clarify "type" is $pr_type in $prestat_u::dir and $prestat_dir


igrep:
Let me file an issue beforehand to make it sure (I've already joined in the Community Group. So I'm ready to send a PR if okay).

I suggest the word "type" in the descriptions of [`$prestat_u::dir`](https://github.com/WebAssembly/WASI/blob/c8517544c913e1c9646be4cdfb3fcc687412c83e/phases/ephemeral/witx/typenames.witx#L697) and [`$prestat_dir`](https://github.com/WebAssembly/WASI/blob/c8517544c913e1c9646be4cdfb3fcc687412c83e/phases/ephemeral/witx/typenames.witx#L686) should be clarified as `$pr_type`.
We can actually easily infer from the context, but I felt confused a little at first.

pchickey:
Thanks. We welcome all PRs that make the docs clearer. That particular text is going away with #220, but I'm happy to accept a PR in the meantime.

PR #222, opened 2020-02-06
Clarify "type" is $pr_type in $prestat_u::dir and $prestat_dir

igrep:
The word "type" in the descriptions of `$prestat_u::dir` and `$prestat_dir` should be clarified as `$pr_type`.
We can actually easily infer from the context, but I felt confused a little at first.

Fix: https://github.com/WebAssembly/WASI/issues/221


sunfishcode:
Looks good to me too.

Issue #174, opened 2019-12-16
Framebuffer API


olanod:
What do you think of my comment in https://github.com/WebAssembly/WASI/issues/171#issuecomment-566061984. Feasible? Not expert in the subject so I might not see lots of small details.

> ...
> I'd like to see a frame buffer API some time, for embedded contexts for example a buffer that I fill with some pixel information(e.g. a micro writing B&W pixels of a small e-ink display) is all that would be needed I guess. The frame buffer can be a wasi file descriptor, perhaps the user can request/create more than one buffer or is given one by the host, then with existing APIs something like window resizing could be detected by listening to [events on the file descriptor](https://github.com/bytecodealliance/wasmtime/blob/master/docs/WASI-api.md#__wasi_eventtype_t-uint8_t), a __WASI_EVENTTYPE_FD_READ could signal that the buffer changed(window resized?) a __WASI_EVENTTYPE_FD_WRITE could be the requestAnimationFrame of WASI world.
> Once the user renders pixels on the buffer by hand or with a graphics API like webGPU then the host decides what to do with that, could be a headless set up where the frame-buffer is dumped to a file or telling the compositor of the platform hey here are some pixels!, put it in the window you gave me!(window was created by the run-time host, not the user?).

programmerjake:
I think having a method to access a memory-mapped frame buffer (not necessarily in GPU memory, can just be system memory) is a good idea -- it exists in all of (non-exhaustive list) Win32, X11, Wayland, KMS/DRM (Linux and BSDs), and nearly all embedded devices with a display and can be easily emulated on other systems.

This allows for fast software rendering when desired, including on GPU-less devices.

This is not a replacement for GPU APIs such as WebGL or WebGPU, which we should also support (except where there is no GPU, such as small embedded systems), but each method should get exclusive access to each Vulkan Surface (or equivalent) to avoid needing complex synchronization.

pierogitus:
Perhaps the baseline should be a compositor/animation API. You create a surface then animate it with its handle and a matrix (similar to CSS transform). Animating could be a capability right distinct from read/write. Maybe you could even generate a new handle where the read/write capabilities were constrained to a 2D region of the surface.

Also any graphics API brings up the need for an accessibility API

torch2424:
Hello!

I was thinking about picking this up as a personal project in my free time 😄 

[...]

PR #229, opened 2020-02-24
Started the Framebuffer API - Am I doing things right so far?

torch2424:
relates to #174 

cc @sunfishcode 

# What is this PR?

This starts a simple draw syscall for a possible framebuffer API. **This is a draft PR where I am more just asking if I seem to be on the right track so far with my implementation on how i can contribute this API.**

This (so far) creates: `phases/ephemeral/witx/wasi_ephemeral_framebuffer.witx` which will be the home for all of the syscalls used by the framebuffer API. And this was tested on my fork of [wasmtime on my framebuffer branch](https://github.com/torch2424/wasmtime/commit/b0b69f14603c15257f99d41c2c019f91cc6e140c). 

# What is the proposed Framebuffer API?

Please see #174 for more context. But to me, it will eventually become the equivalent of [writing directly to something like a linux framebuffer](http://seenaburns.com/2018/04/04/writing-to-the-framebuffer/) but standardized in WASI.

# How did I get this up and running to start contributing?

[Josh Triplett's, Extending WASI and wasmtime from WasmSF](https://www.youtube.com/watch?v=9py3lqYoihc) was a great starting point for me! The technical "how-to contribute" is already out of date. But I still think it is great for the context, and still is somewhat true for what you need to do (covered in my how-to below). 

(Per @sunfishcode request) Here's how I was able to import my own WASi syscall to my wasm program, and then run it on my forked wasm runtime:

1. Fork the [WASI](https://github.com/WebAssembly/WASI) repo.
2. Clone your WASI fork.
3. Fork the [wasmtime]() repo (Using wasmtime as it was the one used in Josh Triplett's talk).
4. Clone your wasmtime fork using the build instructions on [wasmtime.dev](https://wasmtime.dev/).
5. In your WASi fork, implement your syscall in a `.witx` file in : `phases/ephemeral/witx` (For example, [here is my initial draw syscall](https://github.com/torch2424/WASI/commit/9512fa34ee06ced3765aa6e2926e94635c8859d9)).
6. In your wasmtime fork, that you correctly made sure to get the submodules for when you cloned it, implement this same syscall in the latest snapshot. In my case it was in: `crates/wig/WASI/phases/snapshot/witx/wasi_snapshot_preview1.witx`. **NOTE:** We are implementing this in the latest snapshot so there is less you have to do to have wasmtime allow it.
7. In your wasmtime fork, implement the actual function that does the logic of your syscall in: `crates/wasi-common/src/hostcalls_impl/misc.rs`. (For example, here is my "println does it work?" [draw syscall implementation](https://github.com/torch2424/wasmtime/commit/b0b69f14603c15257f99d41c2c019f91cc6e140c))
8. Build your wasmtime fork using the instructions in [wasmtime.dev](https://wasmtime.dev/) (NOTE: I had to use Rust nightly).
9. Write a wasm module that calls your WASI syscall (For example, here is my [draw.wat compiled with wat2wasm](https://github.com/torch2424/wasi-syscall-playground/blob/fb1e6511d7f97d7a71306f1e01ec0690a8789da7/draw.wat))
10. Run your compiled wasm module with your compiled wasmtime fork (wasmtime executable should be in `target/release/wasmtime`). For Example: `./target/release/wasmtime ../wasi-syscall-playground/draw.wasm`.
11. Your syscall should be run! :smile: 

# Results

Screenshot from 2020-02-24 01-50-32

EDIT: Meant to make a draft PR, but I clicked the wrong button, [and there is no way to go back](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107/page/5). Closing for now, and I'll keep working on it until it is actually ready and re-open?

### DIFF
--- a/phases/ephemeral/witx/wasi_ephemeral_framebuffer.witx
+++ b/phases/ephemeral/witx/wasi_ephemeral_framebuffer.witx
@@ -1,0 +1,21 @@
+;; WASI Environment Variables.
+;;
+;; This is a `witx` file. See [here](https://github.com/WebAssembly/WASI/tree/master/docs/witx.md)
+;; for an explanation of what that means.
+
+(use "typenames.witx")
+
+(module $wasi_ephemeral_framebuffer
+ ;;; Linear memory to be accessed by WASI functions that need it.
+ (import "memory" (memory))
+
+ ;;; Function to output rgba array to the framebuffer
+ (@interface func (export "draw")
+  ;; Pointer to the buffer
+  (param $frame_buf (@witx pointer u8))
+  ;; Length of the buffer
+  (param $buf_length u64)
+  ;; The error code of the operation
+  (result $error $errno)
+ )
+)
--- a/phases/ephemeral/witx/wasi_ephemeral_framebuffer.witx
+++ b/phases/ephemeral/witx/wasi_ephemeral_framebuffer.witx
@@ -14,7 +14,7 @@
   ;; Pointer to the buffer
   (param $frame_buf (@witx pointer u8))
   ;; Length of the buffer
-  (param $buf_length u64)
+  (param $buf_length $size)
   ;; The error code of the operation
   (result $error $errno)
  )


programmerjake:
I would suggest having an API where you can:
* `map_buffer`: convert a passed-in memory region to a memory-mapped buffer with a particular pixel format and layout kind, which can be implemented by remapping a region of the linear memory to point to the HW video ram if the HW supports that (can use `mmap` to remap a region on Linux), or just leave it as plain system ram. There could also be an option of just returning a pointer to a different address rather than requiring allocated memory to be passed in.

* `unmap_buffer`: tear down the stuff done by `map_buffer`

* `start_update`: start an update to the display with the following parameters:
    * `modified_rect`: a passed-in rectangle as a hint for which region was modified, used to reduce power consumption for partial updates
    * `display_from_location`: the location in the memory-mapped buffer to display from, allows page-flipping and limited scrolling.
* `wait_for_last_update`: waits for the operation initiated by the last `start_update`. not necessary to call `wait_for_last_update` for the display update to complete. This can be used to limit rendering speed to the refresh rate of the display if the updates started by `start_update` wait till the next refresh.

The displayed output may be updated at any time from the last `display_from_location` without any calls to `start_update` (eg. the mapped memory is video ram and whatever is written is displayed right away).
Any displayed output pixels are not guaranteed to be updated after writing to the buffer unless `start_update` is called with the pixels included in `modified_rect`.

There will also need to be functions to get attributes such as:
* output size in pixels
* output size in millimeters (for proper scaling and handling non-square pixels)
* required memory-mapped buffer size and alignment for different pixel formats and number of pages for page-flipping.
* allowed pixel formats
* allowed `display_from_location` values

there is obviously more stuff that I didn't include or overlooked.

programmerjake:
> EDIT: Meant to make a draft PR, but I clicked the wrong button, [and there is no way to go back](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107/page/5). Closing for now, and I'll keep working on it until it is actually ready and re-open?

@torch2424
It would probably be better to reopen but change the title to begin with `WIP: `, since that improves visibility.

PR #234, opened 2020-02-26
Reorder params/results in clock_res_get syscall

kubkon:
This is merely a stylistic nit which strives to maintain consistency
across all WASI syscalls in that `param`s *always* precede `result`s
in `*.witx` specs.


sunfishcode:
Tools should still aim to support intermixed params and returns like this, since the syntax comes from wit, which gets it from wat, which permits that, but it makes sense to keep WASI's witx files tidy.

kubkon:
> Tools should still aim to support intermixed params and returns like this, since the syntax comes from wit, which gets it from wat, which permits that, but it makes sense to keep WASI's witx files tidy.

And I can confirm that `witx` tool handles it just fine :-)

PR #235, opened 2020-02-26
Disallow non-portable and signal values as exit statuses.

sunfishcode:
Exit codes of at least 256 aren't portable to [POSIX exit], so
programs expecting to return full 32-bit [Windows System Error Codes]
aren't practically portable.

And on POSIX, error codes of at least 128 are reserved for reporting
program exits via signals, and 127 and 126 are reserved for POSIX-style
shells. While it's theoretically possible for POSIX applications to
return these explicitly, this is very rare, not often useful,
particularly in programs intended to be portable, and could potentially
be confusing to users.

If a need arrises for programs to return values in [126,256), or to
provide other kinds of information upon program exit, we can look at
relaxing these restrictions or adding new APIs to WASI for program
termination, but for now it makes sense to start with something simple.

With that, this PR proposes:
 - The WASI `exit` function takes a `u8`, but if the value is at
   least 126, it traps. Otherwise it is provided to the environment.
 - WASI libc's `exit` will map from `int` to `u8` by applying the mask
   as specified in [POSIX exit].

No other WASI syscalls trap right now, but `exit` has no other way to
indicate errors.

[POSIX exit]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/_Exit.html
[Windows System Error Codes]: https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes?redirectedfrom=MSDN#system-error-codes


phases/ephemeral/witx/wasi_ephemeral_proc.witx
```diff
@@ -8,9 +8,12 @@
...
 
 (module $wasi_ephemeral_proc
-  ;;; Terminate the process normally. An exit code of 0 indicates successful
-  ;;; termination of the program. The meanings of other values is dependent on
-  ;;; the environment.
+  ;;; Terminate the process normally. An exit code of `$exitcode::success`
+  ;;; reports successful completion of the program. An exit code of
+  ;;; `$exitcode::failure` or any other value less than 126 reports a
+  ;;; failure, and the value is provided to the environment. If a value
+  ;;; of at least 126 is given, this function behaves as if it were
```
sbc100: I find the phrases "If a value of at least 126 is given" I little hard to parse.   Wouldn't "If a value of 126 or above..", or "If a value greater than or equal to 126.." be more a commonly used phasing.

I noticed this phrases on the PR description and was momentarily confused there too.   Is this choice of phrase deliberate, or following some precedent?

sunfishcode: Heh, I'm reading a book right now that uses "at least" and "at most" instead of "greater than or equal" and "less than or equal". It makes sense in that context, but this PR isn't in that context ;-). I'll change it.

peterhuene:
@sunfishcode as Wasmtime is enforcing this restriction for `proc_exit`, should this get merged into the next WASI snapshot?

sunfishcode:
Yes, let's merge this and aim for the next snapshot.

PR #243, opened 2020-03-13
Move `witx` program to an example

alexcrichton:
This allows crates which depend on `witx` for library support to avoid
pulling in `pretty_env_logger` and `structopt` since those are only used
by the binary program instead of the library itself.


alexcrichton:
Nah it's not critical at all, find to wait until the next one.

PR #244, opened 2020-03-13
fix error message when docs out of date

pchickey:
The invocation for the docs generator changed with #243 .  Remembered this right after I merged.

### DIFF
--- a/tools/witx/tests/wasi.rs
+++ b/tools/witx/tests/wasi.rs
@@ -155,7 +155,9 @@ fn diff_against_filesystem(expected: &str, path: &Path) {
     }
 
     eprintln!();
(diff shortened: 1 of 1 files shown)


tools/witx/tests/wasi.rs
```diff
@@ -155,7 +155,9 @@ fn diff_against_filesystem(expected: &str, path: &Path) {
     }
 
     eprintln!();
-    eprintln!("To regenerate the files, run `cd tools/witx && cargo run -- repo-docs`.");
+    eprintln!(
+        "To regenerate the files, run `cd tools/witx && cargo run -- --example witx repo-docs`."
```
alexcrichton: I think the `--` will need to be removed, but otherwise looks good!

pchickey: thanks!

Issue #242, opened 2020-03-12
[Wasi Meetings]: Change Meeting Time in Accordance with Daylight Savings


torch2424:
Hello!

My apologies for missing the meeting today. I went off the UTC time listed on the meeting page: https://github.com/WebAssembly/WASI/blob/master/meetings/2020/WASI-03-12.md

UTC 17:00 is 10:00AM US PST. I assume this was an issue due to daylight savings in the states. We had a similar problem at AssemblyScript, and are adjusting our time because of it :smile: 

Hope this helps!

cc @sunfishcode since they host the meetings :smile: :+1:

sunfishcode:
Oops! Yes, you're exactly right; thanks for letting me know. I'm sorry you missed the meeting. I'll correct this for future meetings.

sunfishcode:
I've now posted an agenda for the meeting on the 26th in #248, with a corrected time.

PR #248, opened 2020-03-24
Add an agenda for the 03-26 meeting.

sunfishcode:
And add the notes for the 03-12 meeting.


sunfishcode:
I've also updated the time in the meeting agenda to hopefully fix https://github.com/WebAssembly/WASI/issues/242.

PR #250, opened 2020-03-28
witx: expose iterator of all error types

pchickey:
In the wiggle code generator, we require that the error types of a given Document implement certain traits. This method iterates over all of the unique types that are used as error types in a Document.

I'll release this as 0.8.5 once its merged.

### DIFF
--- a/tools/witx/src/ast.rs
+++ b/tools/witx/src/ast.rs
@@ -1,5 +1,5 @@
 #![allow(dead_code)]
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::rc::{Rc, Weak};
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


alexcrichton:
Should this perform a small check to see if the first type is named `error` or `errno`? It seems plausible that one day APIs may be added that could fail and in that case the first argument might not necessarily be an error?

pchickey:
That is a good question. Right now we just have a convention, and its fragile and surprising whether you base it on the name (errno, error, anything with those as a substring, etc) or the position in the result set.

I think we should instead extend or change the witx language to explicitly indicate which result in a given function is for reporting errors. We could define a new field named e.g. `error-result`, or use an annotation like e.g. `(@witx error)` in place of the name identifier in a `(result name type)`, or maybe something else I haven't thought of.

pchickey:
Yeah, I want to get to that change at some point, but it hasn't seemed urgent as long as the library of witx files in the wild is small enough that we can make a PR to update them all. Thankfully, we are not too successful yet.

PR #252, opened 2020-04-03
Start moving WASI to a CG-style phases process.

sunfishcode:
Adapt the WebAssembly CG's phases process for WASI, and make an
initial mapping of current WASI proposals to phases within this
process. This is quite rough at this time, and not everything
fits in quite yet, and we can adjust things. This is a first
step, and we'll work for greater harmonization as we proceed.


sunfishcode:
Merging, as discussed at the meeting last week. We can continue to iterate on the process from here!

PR #256, opened 2020-04-13
Describe a primitive reactor ABI.

sunfishcode:
This adds a new reactor ABI, indicated by exporting the function
"_activate". For now, a reactor is just a module which doesn't have
a lifetime scoped by a "_start" or "main" function, though it may
evolve to include other semantics in the future.

### DIFF
--- a/design/application-abi.md
+++ b/design/application-abi.md
@@ -20,7 +20,7 @@ There are two kinds of modules:
    is considered terminated and no further use of any of its exports should be
    permitted.
 
- - A *reactor* exports a function named `_activate`, with no arguments and no
+ - A *reactor* exports a function named `_initialize`, with no arguments and no
    return values. Environments shall call this function once, after instantiating
    the module and all of its dependencies. After this function exits, the instance
    remains live, and its exports may be accessed.
--- a/design/application-abi.md
+++ b/design/application-abi.md
@@ -12,20 +12,25 @@ include.
 Current Unstable ABI
 --------------------
 
-The current WASI unstable ABI specifies only two exports from a WASI
-application:
+There are two kinds of modules:
 
-- `_start` - the program entry point
-- `memory` - linear memory used by the program
+ - A *command* exports a function named `_start`, with no arguments and no return
+   values. Environments shall call this function once, after instantiating the
+   module and all of its dependencies. After this function exits, the instance
+   is considered terminated and no further use of any of its exports should be
+   permitted.
 
-The `_start` export must be WebAssembly function and will be used as the program
-entry point.  This is the default name used by `lld` when linking WebAssembly
-modules.  The embedder is expected to call this function once the module is
-instantiated.
+ - A *reactor* exports a function named `_activate`, with no arguments and no
+   return values. Environments shall call this function once, after instantiating
+   the module and all of its dependencies. After this function exits, the instance
+   remains live, and its exports may be accessed.
 
-Many of the current WASI unstable APIs require sharing of linear memory between
-the application and the embedder.  In order to use any such APIs the WASI module
-is expected to export its linear memory under the name `memory`.
+Regardless of the kind, all programs accessing WASI APIs also export a linear
+memory with the name `memory`. Pointers in WASI API calls are relative to this
+memory's index space.
+
+In the future, as the underlying WebAssembly platform offers more features, we
+we hope to eliminate the requirement to export all of linear memory.
 
 Planned Stable ABI
 ------------------
--- a/design/application-abi.md
+++ b/design/application-abi.md
@@ -25,6 +25,10 @@ There are two kinds of modules:
    the module and all of its dependencies. After this function exits, the instance
    remains live, and its exports may be accessed.
 
+These kinds are mutually exclusive; implementations should report an error if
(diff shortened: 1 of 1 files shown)


sunfishcode:
I should also mention, a lot of people are already effectively using reactor-style modules in practice. This change just allows us to support this use case properly, as it allows the C++ and libc runtimes to initialize themselves.

design/application-abi.md
```diff
@@ -12,20 +12,25 @@ include.
...
+   values. Environments shall call this function once, after instantiating the
+   module and all of its dependencies. After this function exits, the instance
+   is considered terminated and no further use of any of its exports should be
+   permitted.
 
-The `_start` export must be WebAssembly function and will be used as the program
-entry point.  This is the default name used by `lld` when linking WebAssembly
-modules.  The embedder is expected to call this function once the module is
-instantiated.
+ - A *reactor* exports a function named `_activate`, with no arguments and no
```
sbc100, line 23: Can you explain why you went with `_activate` rather then something more conventional like `_init` or `_initialize`?

design/application-abi.md
```diff
@@ -12,20 +12,25 @@ include.
...
+   return values. Environments shall call this function once, after instantiating
+   the module and all of its dependencies. After this function exits, the instance
+   remains live, and its exports may be accessed.
 
-Many of the current WASI unstable APIs require sharing of linear memory between
-the application and the embedder.  In order to use any such APIs the WASI module
-is expected to export its linear memory under the name `memory`.
+Regardless of the kind, all programs accessing WASI APIs also export a linear
+memory with the name `memory`. Pointers in WASI API calls are relative to this
+memory's index space.
```
sbc100, line 30: Should we say anything about the wasm table here?  Should the module also export its table?  (Unrelated to this particialar PR I suppose).

design/application-abi.md
sunfishcode: In conventional toolchains, `_init` is used in both commands and dynamic libraries. I was going for something that would connote that the presence of this export is what determines whether the thing is a command or an export.

design/application-abi.md
sunfishcode: The main reason we'd want to export the table would be to allow syscalls to call function pointers in it, and we don't currently have any APIs with callbacks. We may in the future, but it's also possible that by the time we do, we'll have better ways to do it than just exporting the whole table. So my sense is to wait for now.

sbc100: Now that we have reactors people can export reactor functions that return function pointers if they want, right?  I guess we can wait for ppl to ask for it.

sunfishcode: Within the context of WASI here, yeah, we can figure that out when we have a proposal that needs it.

design/application-abi.md
sbc100: I think its reasonable to call this `_init`.   Its presence as a export can signal the application is a reactor.

Should we make these mutually exclusive too, such that exporting both these symbols is invalid?

sunfishcode: Good point about mutual exclusion. I've now added a sentence about that.

Concerning the name `_init`, I'm not attached to `_activate`, but I am nervous about having a thing named `_init` which is similar but subtly different from [the ELF `_init`](https://git.musl-libc.org/cgit/musl/tree/crt/crt1.c#n9) in several respects. The ELF `_init` isn't what gets called from the outside, it is used in executables as well as dynamic libraries, it's no longer used to run C++ constructors or libc init in most systems, and it can be awkwardly extended by end users through defining `.init` sections, and none of these apply to the `_activate` function I'm proposing.

sbc100: Still its is a better name IMO.    How about `_initialize` since it avoids that name clash and the needless contraction.

design/application-abi.md
sbc100: OK SGTM.

design/application-abi.md
sunfishcode: Works for me. I've now updated the PR.

design/application-abi.md
```diff
@@ -12,20 +12,29 @@ include.
 Current Unstable ABI
 --------------------
 
-The current WASI unstable ABI specifies only two exports from a WASI
-application:
+There are two kinds of modules:
 
-- `_start` - the program entry point
-- `memory` - linear memory used by the program
+ - A *command* exports a function named `_start`, with no arguments and no return
```
lachlansneff, line 17: Partially off topic, but why not pass some configuration data to `_start`?

For example, the function could receive the command line args, the allowed paths, etc. These would be through interface types, of course.

sunfishcode: Yes, exactly. There are a lot of things we could do here, but we need interface types to do them well. The current `argv`/`environ` interfaces we have now aren't super amazing, and `_start` isn't a great name, and so on, but at the moment it all works well enough that it doesn't seem worth breaking compatibility until we have a big reason for it, like interface types.

sunfishcode:
As I mentioned [in this comment](https://github.com/WebAssembly/WASI/pull/256#discussion_r410337108), this ABI is expected to change in the future, so this won't be the final word here, but it is simple solution to a present problem, so let's move forward.

PR #257, opened 2020-04-15
Document atomicity of `write` and `pwrite`, and technical fix about POSIX

igrep:
See the commit message of each commit for details.

### DIFF
--- a/phases/ephemeral/docs.md
+++ b/phases/ephemeral/docs.md
@@ -1466,10 +1466,9 @@ A buffer into which to write the preopened directory name.
 Write to a file descriptor, without using and updating the file descriptor's offset.
 Note: This is similar to `pwritev` in Linux (and other Unix-es).
 
-Like Linux (and other Unix-es), [`pwrite`](#pwrite) should write all data in the `$fd` is
-for a regular file or a symbolic link to a regular file: Any calls of [`pwrite`](#pwrite)
-(and other functions to read or write on the file) by other threads in the WASI
-process should not be interleaved while [`pwrite`](#pwrite) is executed.
+Like Linux (and other Unix-es), Any calls of [`pwrite`](#pwrite) (and other
+functions to read or write) for a regular file by other threads in the
+WASI process should not be interleaved while [`pwrite`](#pwrite) is executed.
 
 ##### Params
 - <a href="#pwrite.fd" name="pwrite.fd"></a> `fd`: [`fd`](#fd)
@@ -1616,10 +1615,9 @@ The current offset of the file descriptor, relative to the start of the file.
 Write to a file descriptor.
 Note: This is similar to `writev` in POSIX.
 
-Like POSIX, [`write`](#write) should write all data in `$iovs` atomically if the `$fd` is
-for a regular file or a symbolic link to a regular file: Any calls of [`write`](#write)
-(and other functions to read or write on the file) by other threads in the WASI
-process should not be interleaved while [`write`](#write) is executed.
+Like POSIX, Any calls of [`write`](#write) (and other functions to read or write)
+for a regular file by other threads in the WASI process should not be
+interleaved while [`write`](#write) is executed.
 
 ##### Params
 - <a href="#write.fd" name="write.fd"></a> `fd`: [`fd`](#fd)
--- a/phases/ephemeral/docs.md
+++ b/phases/ephemeral/docs.md
@@ -1410,7 +1410,7 @@ The permissions associated with the file.
 
 #### <a href="#pread" name="pread"></a> `pread(fd: fd, iovs: iovec_array, offset: filesize) -> (errno, size)`
 Read from a file descriptor, without using and updating the file descriptor's offset.
-Note: This is similar to `preadv` in POSIX.
+Note: This is similar to `preadv` in Linux (and other Unix-es).
 
 ##### Params
 - <a href="#pread.fd" name="pread.fd"></a> `fd`: [`fd`](#fd)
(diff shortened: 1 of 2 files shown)


programmerjake:
Is it technically possible to provide atomic writes with respect to non-WASI code, such as other processes? (Seems unlikely when a file is `mmap`ed.) If not, wording should be added to limit the scope of the atomicity to the WASI process (implementable with a mutex or similar) or some other suitable scope.

igrep:
Thanks for pointing out. Reviewing [the related chapter of POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_09_07), it refers only to threads.
I'll fix.

sunfishcode:
The related chapter in POSIX only provides these guarantees when the functions are operating on regular files or symbolic links. Unless we intend WASI to differ, we should preserve that aspect of the guarantee.

`write` etc. don't have a guarantee to write all the data or nothing. POSIX only says these "functions shall be atomic with respect to each other", which I read as saying that these functions' I/O won't be interleaved with other I/O, however they can presumably still produce incomplete results if they encounter errors such as running out of space on an output device. And in practice, some OS's also impose their own limitations -- on 64-bit Linux for example, calling `write` with a 4 GiB buffer (which is below `SSIZE_MAX`) to a normal file writes only 0x7ffff000 bytes.

igrep:
> The related chapter in POSIX only provides these guarantees when the functions are operating on regular files or symbolic links

Thanks! I missed the phrase "when they operate on regular files or symbolic links".

> write etc. don't have a guarantee to write all the data or nothing. POSIX only says these "functions shall be atomic with respect to each other", which I read as saying that these functions' I/O won't be interleaved with other I/O, however they can presumably still produce incomplete results if they encounter errors such as running out of space on an output device

I agree. So I rewrote about what `atomicicity` stands for.

https://github.com/WebAssembly/WASI/pull/257/commits/13ab34bff433b422bbe267bfb78f6d776c45594c

phases/ephemeral/docs.md
```diff
@@ -1190,7 +1190,12 @@ A buffer into which to write the preopened directory name.
 
 #### <a href="#pwrite" name="pwrite"></a> `pwrite(fd: fd, iovs: ciovec_array, offset: filesize) -> (errno, size)`
 Write to a file descriptor, without using and updating the file descriptor's offset.
-Note: This is similar to `pwritev` in POSIX.
+Note: This is similar to `pwritev` in Linux (and other Unix-es).
+
+Like Linux (and other Unix-es), [`pwrite`](#pwrite) should write all data in the `$fd` is
+for a regular file or a symbolic link to a regular file: Any calls of [`pwrite`](#pwrite)
```
sunfishcode, line 1196: Is `in` meant to be `if` here?

Also, POSIX doesn't say that `pwrite` etc. will write *all* the data; they could still be interrupted in the middle if an error occurs, such running out of space on the output device.

Also, as a minor correction, POSIX resolves symbolic links at `open` time, so it's not necessary to call out symblic links here in `pread` and `pwrite`. The reason POSIX mentions symbolic links in the paragraph linked above is that the list of affected functions includes `readlink` and `symlink` which operate on symbolic links themselves.

phases/ephemeral/docs.md
```diff
@@ -1464,7 +1464,11 @@ A buffer into which to write the preopened directory name.
 
 #### <a href="#pwrite" name="pwrite"></a> `pwrite(fd: fd, iovs: ciovec_array, offset: filesize) -> (errno, size)`
 Write to a file descriptor, without using and updating the file descriptor's offset.
-Note: This is similar to `pwritev` in POSIX.
+Note: This is similar to `pwritev` in Linux (and other Unix-es).
+
+Like Linux (and other Unix-es), Any calls of [`pwrite`](#pwrite) (and other
```
MaxDesiatov, line 1469:
```suggestion
Like Linux (and other Unix-es), any calls of [`pwrite`](#pwrite) (and other
```

phases/ephemeral/docs.md
```diff
@@ -1611,6 +1615,10 @@ The current offset of the file descriptor, relative to the start of the file.
 Write to a file descriptor.
 Note: This is similar to `writev` in POSIX.
 
+Like POSIX, Any calls of [`write`](#write) (and other functions to read or write)
```
MaxDesiatov, line 1618:
```suggestion
Like POSIX, any calls of [`write`](#write) (and other functions to read or write)
```

phases/ephemeral/witx/wasi_ephemeral_fd.witx
```diff
@@ -157,7 +157,11 @@
   )
 
   ;;; Write to a file descriptor, without using and updating the file descriptor's offset.
-  ;;; Note: This is similar to `pwritev` in POSIX.
+  ;;; Note: This is similar to `pwritev` in Linux (and other Unix-es).
+  ;;;
+  ;;; Like Linux (and other Unix-es), Any calls of `pwrite` (and other
```
MaxDesiatov, line 162:
```suggestion
  ;;; Like Linux (and other Unix-es), any calls of `pwrite` (and other
```

phases/ephemeral/witx/wasi_ephemeral_fd.witx
```diff
@@ -250,6 +254,10 @@
 
   ;;; Write to a file descriptor.
   ;;; Note: This is similar to `writev` in POSIX.
+  ;;;
+  ;;; Like POSIX, Any calls of `write` (and other functions to read or write)
+  ;;; for a regular file by other threads in the WASI process should not be
+  ;;; interleaved while `write` is executed.
```
MaxDesiatov, line 260:
```suggestion
  ;;; Like POSIX, any calls of `write` (and other functions to read or write)
  ;;; for a regular file by other threads in the WASI process should not be
  ;;; interleaved while `write` is executed.
```

igrep:
@MaxDesiatov Thank you for the review!

sunfishcode:
This looks good to me, thanks for seeing this through!

PR #264, opened 2020-04-27
Move $d_namlen to the end of $dirent

caspervonb:
Just a minor (although ABI breaking) nit suggestion that stuck out to me when implementing `fd_readdir`.

Currently `$dirent` has the filetype in-between the name length and the name data so you end up scanning the length of entry's name, writing the type of the entry then going back to write out the name data.

### DIFF
--- a/phases/ephemeral/docs.md
+++ b/phases/ephemeral/docs.md
@@ -558,13 +558,13 @@ The serial number of the file referred to by this directory entry.
 
 Offset: 8
 
-- <a href="#dirent.d_namlen" name="dirent.d_namlen"></a> `d_namlen`: [`dirnamlen`](#dirnamlen)
-The length of the name of the directory entry.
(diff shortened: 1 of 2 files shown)


sunfishcode:
I agree it is nicer to keep the name length next to the name data. Would you mind also updating phases/ephemeral/docs.md, either by hand or by running `cd tools/witx && cargo run --example witx repo-docs`? Thanks!

caspervonb:
Done and force pushed https://github.com/WebAssembly/WASI/commit/2f64d16f2d3b4d55f8230ceab6c0737f4f9e3f1d

sunfishcode:
My apologies, but the patch to land Size and Alignment documentation landed before this, so there's now a merge conflict. Would you mind resolving this? Thanks!

caspervonb:
Rebased against current master and ran the doc tool again.

sunfishcode:
Thanks!

PR #268, opened 2020-05-01
Added sizes and alignments to WASI docs

### LINKED ISSUE #266: Struct sizes of event and subscription
We've recently run into problems regarding the layout or more precisely sizes of the `event` and `subscription` structs, and I'd like to have some clarification on the following aspects:

* According to [the relevant paragraphs in docs.md](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-event-struct), the `event` struct will always contain the `event_fd_readwrite` struct as by extension, so an `event` struct is always exactly 32 bytes, regardless of event type.

* According to [the relevant paragraphs in docs.md](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-subscription_u-union), the `subscription` struct will always reserve the size necessary to store the largest variant of the `subscription_u` union, so a `subscription` struct is always exactly 48 bytes, regardless of subscription type.

* The size of a `prestat` is always exactly 8 bytes in 32-bit, and 16 bytes in 64-bit.

Are these assumptions correct? If not, could you point me into the right direction? If correct, do you think it'd be useful to add the sizes to the documentation, so other language implementers trying to model these where there's no native concept of C structs don't run into similar problems in the future?

torch2424:
closes #266 

This adds documentation for all the sizes, alignments, and offset of data types, structs, and unions for `.witx` documents. :smile: :+1: 

cc @pchickey for the review :smile: 

# Screenshot

*I was using a local md to html converter to view the doc, which is why it looks a little funny :smile: *

Screenshot from 2020-04-30 17-37-08

### DIFF
--- a/tools/witx/src/docs/ast.rs
+++ b/tools/witx/src/docs/ast.rs
@@ -10,6 +10,7 @@ use crate::{
     },
     polyfill::{FuncPolyfill, ModulePolyfill, ParamPolyfill, Polyfill, TypePolyfill},
     RepEquality,
+    layout::Layout,
 };
 
 fn heading_from_node(node: &MdNodeRef, levels_down: usize) -> MdHeading {
@@ -26,7 +27,7 @@ impl ToMarkdown for Document {
                 heading.new_level_down(),
                 name,
                 name,
-                &d.docs,
+                format!("{}Size: {}\n Alignment: {}", &d.docs, &d.mem_size(), &d.mem_align()).as_str(),
             ));
             d.generate(child.clone());
         }
@@ -173,7 +174,9 @@ impl ToMarkdown for StructDatatype {
         let heading = heading_from_node(&node, 1);
         node.new_child(MdSection::new(heading, "Struct members"));
 
-        for member in &self.members {
+        for member_layout in &self.member_layout() {
+            let member = member_layout.member;
+            let offset = member_layout.offset;
             let name = member.name.as_str();
             let id = if let Some(id) = node.any_ref().id() {
                 format!("{}.{}", id, name)
@@ -184,7 +187,7 @@ impl ToMarkdown for StructDatatype {
                 MdHeading::new_bullet(),
                 id.as_str(),
                 name,
-                &member.docs,
+                format!("{}Offset: {}", &member.docs, &offset).as_str(),
             ));
             member.tref.generate(n.clone());
         }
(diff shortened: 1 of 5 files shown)


tools/witx/src/docs/ast.rs
```diff
@@ -195,8 +204,44 @@ impl ToMarkdown for StructDatatype {
 
 impl ToMarkdown for UnionDatatype {
     fn generate(&self, node: MdNodeRef) {
-        let heading = heading_from_node(&node, 1);
-        node.new_child(MdSection::new(heading, "Union variants"));
+        // Sizes & Alignments
+        let sizes_heading = heading_from_node(&node, 1);
+        node.new_child(MdSection::new(sizes_heading, "Union Layout"));
+        let union_layout = &self.union_layout();
+        node.new_child(MdNamedType::new(
```
kubkon, line 211: If you used a generic `MdSection` instead of an `MdNamedType`, then it won’t be HTML tagged by default.

torch2424:
@kubkon Thank you very much for the review! :smile: :+1: 

I went ahead and made your requested changes. This should be good to go :smile: 

Screenshot from 2020-05-01 11-50-37

kubkon:
On second thought, I'll wait and see what @sunfishcode thinks since it does affect the docs, so I'm not sure whether it requires chairs approval or not ;-)

tools/witx/src/docs/ast.rs
```diff
@@ -26,7 +27,13 @@ impl ToMarkdown for Document {
                 heading.new_level_down(),
                 name,
                 name,
-                &d.docs,
+                format!(
+                    "{}Size: {}\n Alignment: {}",
```
sunfishcode, line 31: Is this space before the "Alignment:" intended? It seems to make the spacing inconsistent with "Size:" and "Offset:".

tools/witx/src/docs/ast.rs
```diff
@@ -184,7 +193,7 @@ impl ToMarkdown for StructDatatype {
                 MdHeading::new_bullet(),
                 id.as_str(),
                 name,
-                &member.docs,
+                format!("{}Offset: {}", &member.docs, &offset).as_str(),
```
sunfishcode, line 196: Would it make sense to add a newline after the offset here? Before this change, it looks like there's a blank line at that point, which creates a nice visual separation of different documentation items.

tools/witx/src/docs/ast.rs
sunfishcode, line 31: Also, similar to the "Offset:", should there be a trailing newline here?

torch2424:
@sunfishcode @kubkon Thanks for the review / approval! :smile: 

@sunfishcode Nice catch on the whitespace stuff! My hack-y markdown to html thing made the Whitespace look a lot different than what Github was rendering. Fixing these now :smile:

torch2424:
@sunfishcode @kubkon 

Found a whitespace I think that looks good in Github. Let me know what you think :smile: 

Screenshot from 2020-05-01 15-59-53

PR #274, opened 2020-05-10
Fix links to dirent struct in fd_readdir description

kubkon:
This commit fixes links in `fd_readdir` description/docs which were
using an outdated (now invalid) ref syntax when referencing `dirent`
struct.

The changes apply to docs only and only to `wasi_unstable` and
`wasi_snapshot_preview1` snapshots. The `wasi_ephemeral` has properly
referenced links.


sunfishcode:
Looks good!

PR #275, opened 2020-05-12
Fix link to event struct

kubkon:
This commit fixes an outdated ref syntax when referencing back to
`event` struct.


sunfishcode:
Looks good!

PR #288, opened 2020-06-11
Describe legacy commands.

sunfishcode:
The core idea of legacy commands is that all function exports are
command entrypoints, and they expect to be called on a fresh instance
which lives only for the direction of the call.

This is a generalization of the existing `_start` convention, which
already has the expectation of a fresh instance  which lives only for
the direction of the call, and is one of the pieces need to enable
user-defined command entrypoints which don't take string arguments and
don't return an i32 status code.

To show how this works in practice, the following patches implement
this new behavior in wasm-ld and wasi-libc:
 - https://reviews.llvm.org/D81689
 - https://github.com/WebAssembly/wasi-libc/pull/203

### DIFF
--- a/legacy/application-abi.md
+++ b/legacy/application-abi.md
@@ -17,6 +17,14 @@ There are two kinds of modules:
  - A *command* exports a function named `_start`, with no arguments and no return
    values.
 
+   `_start` is the default export which is called when the user doesn't select a
+   specific function to call. Commands may also export additional functions,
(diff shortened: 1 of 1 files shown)


sunfishcode:
This adds documentation for what is now the legacy application API. This documents behavior already implemented in clang and wasi-sdk, so my inclination is to merge it to document the current code. Any objections?

PR #311, opened 2020-08-17
latest wast in witx

pchickey:
wast dep from `20.0.0` => `22.0.0`
witx version from `0.8.6` => `0.8.7`


pchickey:
published. thanks!

PR #317, opened 2020-09-05
Remove reference to non existent modules.md file

andrewdavidmackenzie:
I couldn't find a modules.md anywhere else in this repo to link to, so just removed the broken reference.


sunfishcode:
Thanks!

PR #318, opened 2020-09-06
Create witx-type-representation.md

SirJosh3917:
This document summarizes the types that can be found in the phaser snapshots. The goal of this is that it should be a one-stop shop for people looking to implement ABI compatibility with certain types.

Points of improvement before merging:
- More condensed representation of whole bytes? `subscription_u` would be hard to read on mobile
- Pointer<T> needs more elaboration
- Are the images exactly accurate to the positions of bytes?
- Will it answer all ABI protocol related questions? Are there odd edge cases?


design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+# Pointer<T> and Array<T>
+A `Pointer<T>` and `Array<T>` are both just `Pointer<T>`s. A `Pointer<T>`'s size is guaranteed to be 8 bytes, but if on a 32 bit architecture it will only be 4 bytes. The data stored in one of these types is exactly enough to point to some data in RAM. When the pointer is dereferenced, the data stored at that location will be a contiguous amount of `T`s. // how to get length? is there one for ArrayT?
+
+```
+Pointer<u8>
+
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+| 0 0 0 0 0 1 0 0 | 0 0 1 0 1 1 0 0 | 0 0 0 1 1 1 0 0 | 0 0 1 1 0 1 0 0 | 1 0 0 1 0 0 0 0 | 0 0 0 0 0 1 0 0 | 0 0 0 0 1 0 0 1 | 0 0 0 1 0 0 0 0 |
```
sunfishcode, line 83: What do the `1`s here mean?

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
+# Witx Type Representation
+In Witx, each type corresponds to a lower-level type. For example, the `size` type resolves to a `u32`, the `errno` type resolves to an `Enum(u16)`, and more. Detailed below is each type, and what they resolve to in memory.
+
+# u8/u16/u32/u64/s64
+These are C-style `uint8_t`s, `uint16_t`s, `uint32_t`s, `uint64_t`s, and `int64_t`s. In the .NET world, they may be more recognizable as an `byte` (in C# this is an unsigned `byte` if you're coming from Java), `ushort`, `uint`, `ulong`, and `long`. These are typically little-endian, although it may not be necessary to know that.
```
sunfishcode, line 5: It may be worth pointing to https://github.com/WebAssembly/interface-types/blob/master/proposals/interface-types/Explainer.md#integers as the most authoritative reference for the signedness-aware integer types we have right now.

> in C# this is an unsigned `byte` if you're coming from Java

Is this parenthetical about C# or Java?

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
...
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+| 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 |
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+| 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 |
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+```
+
+# Pointer<T> and Array<T>
+A `Pointer<T>` and `Array<T>` are both just `Pointer<T>`s. A `Pointer<T>`'s size is guaranteed to be 8 bytes, but if on a 32 bit architecture it will only be 4 bytes. The data stored in one of these types is exactly enough to point to some data in RAM. When the pointer is dereferenced, the data stored at that location will be a contiguous amount of `T`s. // how to get length? is there one for ArrayT?
```
sunfishcode, line 77: > A `Pointer<T>`'s size is guaranteed to be 8 bytes, but if on a 32 bit architecture it will only be 4 bytes.

This sentence seems to contradict itself.

Pointers are just raw pointers that don't have an inherent length. Arrays have a length, and the way it's stored depends on the context. When an Array type appears as a function argument, it's often represented as two wasm arguments, a pointer (linear memory offset) and a length.

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
...
+
+```
+errno: Enum(u32)
++-----------------+-----------------+-----------------+-----------------+
+| 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 1 0 1 0 1 |
++-----------------+-----------------+-----------------+-----------------+
+^ fault
+```
+
+(`clockid` despite only representing 4 values is an `Enum(u32)`. This is for ABI compatibility.)
```
sunfishcode, line 18: In addition to ABI compatibility, the use of wider types than are needed is often for future-proofing too, so that we can add new values to the enum without having to change its size.

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
...
++-----------------+-----------------+
+  ^ ^ ^ ^
+  | | | trunc
+  | | excl
+  | directory
+  creat
+```
+
+# Struct
+A `Struct` is a type that takes up some contiguous amount of blocks in memory, with each field taking up a specific amount of reserved bytes. Interpreting the bytes as one of the types in Witx will yield a usable value.
```
sunfishcode, line 37: "blocks" here is not a typical word to describe memory. I'd suggest saying just "some contiguous amount of memory".

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
...
+
+cont. 32 bytes for the union's data
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+| 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 |
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+| 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 |
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+| 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 | 0 0 0 0 0 0 0 0 |
++-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
+```
```
sunfishcode, line 74: As you observed, this does get pretty wide. My suggestion here is: the only datatype we have where we think about individual bits is `Flags`. For the rest, just show bytes, rather than bits.

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
+# Witx Type Representation
+In Witx, each type corresponds to a lower-level type. For example, the `size` type resolves to a `u32`, the `errno` type resolves to an `Enum(u16)`, and more. Detailed below is each type, and what they resolve to in memory.
+
+# u8/u16/u32/u64/s64
```
sunfishcode, line 4: Subsections can start with `##` instead of `#`.

sunfishcode (commented):
Cool! This looks like a great start documenting a part of witx which is currently undocumented!

Could you add a link to this from here? https://github.com/WebAssembly/WASI/blob/master/docs/witx.md

One other thing which may be useful to mention in the intro; the representations of these types are exposed when used with linear memory, however witx is also meant to be used in the future in non-linear-memory contexts as well, in which case the representation is opaque.

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
+# Witx Type Representation
+In Witx, each type corresponds to a lower-level type. For example, the `size` type resolves to a `u32`, the `errno` type resolves to an `Enum(u16)`, and more. Detailed below is each type, and what they resolve to in memory.
```
sbc100, line 2: Can you wrap long lines at 80?

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
+# Witx Type Representation
```
sbc100, line 1: Empty line after each heading?

design/witx-type-representation.md
sbc100, line 2: "what they resolve to in memory" seems a little odd.  How about "what the correspond to in the WebAssembly type type system"?

design/witx-type-representation.md
sbc100, line 5: What does `typically` mean here?   Aren't these little endian by definition just like the underlying wasm types?

design/witx-type-representation.md
```diff
@@ -0,0 +1,86 @@
...
+In Witx, each type corresponds to a lower-level type. For example, the `size` type resolves to a `u32`, the `errno` type resolves to an `Enum(u16)`, and more. Detailed below is each type, and what they resolve to in memory.
+
+# u8/u16/u32/u64/s64
+These are C-style `uint8_t`s, `uint16_t`s, `uint32_t`s, `uint64_t`s, and `int64_t`s. In the .NET world, they may be more recognizable as an `byte` (in C# this is an unsigned `byte` if you're coming from Java), `ushort`, `uint`, `ulong`, and `long`. These are typically little-endian, although it may not be necessary to know that.
+
+# Enum(T)
+An `Enum(T)` is, unlike a Rust Enum(T), simply is a `T`. However, the value of T can only be one of the specific variants of the enum. This type lends itself to describing when something can only be one of the enumerations in the group (for example, in a group of Dogs and Cats, you may have an enum representing either a Dog or a Cat).
```
sbc100, line 8: It seems a litte rust-cetric to even mention rust here.  The semantics here a basically equivalent to C++ so if you are going to call out specific language it might make more sense to say `just like in C++ ....`.

design/witx-type-representation.md
sbc100: It probably makes sense to say that pointers themselves are 32-bit does it?    For now pointers in WebAssembly are all 32-bit since wasm32 is the only ABI.

(If you like we could mention that this could be 64bits one day once the spec for wasm64 is done... but I don't think that is necessary today).

sbc100 (commented):
Nice!

I'm sure folks will find this useful. 

I wonder if it makes sense to integrate this into the existing docs somewhere?  Or at least be linked to from witx-generated docs?  @pchickey

design/witx-type-representation.md
SirJosh3917: That was just random uninitialized data for no good reason, replacing all of that with 0x00

design/witx-type-representation.md
SirJosh3917: done & removed that whole parenthetical part, using just the C types is good enough

design/witx-type-representation.md
SirJosh3917: changed contradictory sentence to

> it will only use 4 of the 8 bytes

design/witx-type-representation.md
SirJosh3917: done

design/witx-type-representation.md
SirJosh3917: "WebAssembly type type" system seems like a typo, so I changed that to "WebAssembly type system".

design/witx-type-representation.md
SirJosh3917: yeah you're right - i'll point to [#786](https://github.com/WebAssembly/design/issues/786#issuecomment-244548105) to help drive that point home

design/witx-type-representation.md
SirJosh3917: I had done that because the `Enum(T)` notation looked very similar to rust's notation, and thought that might help clear up misconceptions for people who had the same thought as myself. I'll remove the anecdote entirely.

design/witx-type-representation.md
SirJosh3917: > (If you like we could mention that this could be 64bits one day once the spec for wasm64 is done... but I don't think that is necessary today).

I'd rather mention it now, or it will forever be forgotten.

SirJosh3917:
my bad on taking forever lol

design/witx-type-representation.md
```diff
@@ -0,0 +1,121 @@
...
+
+A `Struct` is a type that takes up some contiguous amount of memory, with each
+field taking up a specific amount of reserved bytes. Interpreting the bytes as
+one of the types in Witx will yield a usable value.
+
+```
+iovec
+
+buf: Pointer<u8> @ offset 0
+buf_len: size @ offset 4
```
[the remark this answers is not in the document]
SirJosh3917: It seems that [the docs don't mention this distinction](https://github.com/WebAssembly/WASI/blob/main/phases/ephemeral/docs.md#-iovec-record)? Should this difference be mentioned even though docs don't mention it?

SirJosh3917:
Apologies for it having been a while since I've worked on this. I tried to update the terminology to be more in-line with what's currently in use, but I might've glossed something over.

vados-cosmonic:
It's been five years and there have been quite a few changes to the repo and WIT/WASI/the ecosystem at large in the meantime -- @monoclex hopefully you don't mind but I'm closing this draft as it probably is unlikely to get finished and merged anytime soon (just using the past as a guide)

Happy to re-open if it's still relevant/you find time though!

Issue #316, opened 2020-09-04
Clarification for implementing Rust-like enums with values, specifically `clockid`


SirJosh3917:
For the [`clockid`](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-clockid-enumu32) enum, it seems to be an `Enum(u32)` with four variants. Is the `u32` used to represent the value of the clock? The size/alignment is 4, but if the clock reports 4 bytes surely there must be another byte to store the variant? Whilst I understand Rust enums conceptually, to me this doesn't make sense at how I'd implement this in a cross platform manner for other languages. Can this be clarified?

pchickey:
Hi Josh, great questions! First off, I'm sorry that witx docs are lacking - the language and tooling has been evolving, and I've been too busy to write any. The short story is, witx types resemble C more than Rust right now, for historical reasons - the original spec of WASI snapshot 0 was a C header, and we wanted to keep ABI compatibility when we changed the canonical description to be witx. We expect this to evolve when the Interface Types spec and toolchain support lands - many of the same people working on WASI are working on that right now.

Witx `enum`s are c-style - they do not have payloads, they are just an integer where only a subset of values (counting up from 0) are valid. The integer type specified in a union declaration determines its representation in memory. So, `clockid` is a `u32` in memory, whereas `errno` is a `u16`. (In the particular case of `clockid`, its wasteful to use 4 bytes to represent 4 variants, but thats what we got stuck with for compatibility.)

Witx `union`s are for rust-style enums that have different types for each variant. The `union` requires a witx `enum` type be provided for use as the discriminant. Unions are laid out like a C `struct { the_variant_enum tag; union { variant_type_1 variant1; ... } u; }`.

If you're using the `witx` crate to parse and validate, the `Layout` trait on AST nodes will provide the size and alignment of any datatype, and there are special `UnionLayout` and `StructMemberLayout` types that describe the interiors of unions and structs.

SirJosh3917:
Thanks for the detailed information! Is there a place I can contribute the information you mentioned (with an obligatory ASCII chart of course) to the docs somewhere? I'd hate for this information to be buried in a Github issue :P

[...]

PR #319, opened 2020-09-06
Add beginnings of a Resources and Handles design document.

monoclex:
File Descriptors seem like they will be a very ubiquitous concept in WASI, thus guiding users to understand them would be extremely helpful. As the implementation is still not fully fleshed out and very much a draft, this document is to remain very incomplete but ideally would have examples and links to other documents so developers can figure out how to use these APIs. This is to resolve another part of #316


design/file-descriptors.md
```diff
@@ -0,0 +1,17 @@
+# File Descriptors
+File Descriptors ([witx type `fd`](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-fd)) are an abstract way of representing access to a resource. More often than not, they are processes (WIP), sockets (WIP), files, or even the console.
+
+# Null File Descriptor
```
caspervonb, line 4: 0 is stdin, not null.

design/file-descriptors.md
```diff
@@ -0,0 +1,17 @@
...
+A File Descriptor of 0 is a null file descriptor. It's used whenever there is *no* file descriptor. Passing this to a method is a no-no, and receiving one from a method must be handled.
+
+# The Console
+When a program is initiated, `stdout`, `stdin`, and `stderr` (respectively `fd` ids `1`, `2`, and `3`) are opened. To print to these, simply call `fd_write` with the corresponding arguments. An simple "Hello World!" program written in `wat` is shown below:
```
caspervonb, line 8: Incorrect.

stdin - 0
stdout - 1
stderr - 2

design/file-descriptors.md
```diff
@@ -0,0 +1,17 @@
...
+
+```wat
+(module
+  ;; TODO: small hello world?
+)
+```
+
+# Files
+Files can be opened and closed with the method [`path_open`](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-path_openfd-fd-dirflags-lookupflags-path-string-oflags-oflags-fs_rights_base-rights-fs_rights_inherting-rights-fdflags-fdflags---errno-fd), and simply drop the returned `fd` to close it. These commands will fail if the WASM environment did not give the program permission to execute these.
```
caspervonb, line 17: This seems to be too specific for a design document as it refers to a specific snapshot and will easily get out of sync.

design/file-descriptors.md
SirJosh3917: Does a null `fd` exist?

design/file-descriptors.md
SirJosh3917: Fixed.

design/file-descriptors.md
SirJosh3917: Would modifying the URL to a less precise URL fix this? If not, does the document have a better fit elsewhere?

design/file-descriptors.md
```diff
@@ -0,0 +1,14 @@
...
+File Descriptors ([witx type `fd`](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-fd)) are an abstract way of representing access to a resource. More often than not, they are processes (WIP), sockets (WIP), files, or even the console.
+
+# The Console
+When a program is initiated, `stdout`, `stdin`, and `stderr` (respectively `fd` ids `0`, `1`, and `2`) are opened. To print to these, simply call `fd_write` with the corresponding arguments. An simple "Hello World!" program written in `wat` is shown below:
+
+```wat
+(module
+  ;; TODO: small hello world?
+)
+```
```
sunfishcode, line 11: Wasmtime's testsuite has [a hello world test](https://github.com/bytecodealliance/wasmtime/blob/main/tests/wasm/hello_wasi_snapshot1.wat) which you can base this on.

It's admittedly a little inconvenient to set up the iovec structures. The idea of adding more convenient non-vectored `read` and `write` functions to WASI has come up a few times; I'd be interested to hear if to think that'd be worth doing.

design/file-descriptors.md
```diff
@@ -0,0 +1,14 @@
+# File Descriptors
+File Descriptors ([witx type `fd`](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-fd)) are an abstract way of representing access to a resource. More often than not, they are processes (WIP), sockets (WIP), files, or even the console.
+
+# The Console
```
sunfishcode, line 4: As a minor style nit, when there are multiple sections, use `##` or so, as `#` is meant for titles.

design/file-descriptors.md
```diff
@@ -0,0 +1,14 @@
+# File Descriptors
+File Descriptors ([witx type `fd`](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-fd)) are an abstract way of representing access to a resource. More often than not, they are processes (WIP), sockets (WIP), files, or even the console.
```
sunfishcode, line 2: This is a good start, and we can build from here. You're right that file descriptors are a foundational concept, and I expect over time we'll add on more text here. One thing that would be useful to say right now is that "file descriptors" are also sometimes called "handles", as we tend to use those terms somewhat interchangeably today.

sunfishcode (commented):
I apologize for being slow here! This is a great topic to start documenting, as I agree, file descriptors are one of the most important concepts we have. This document looks like a good start -- I expect a lot of things will evolve over time, but it's useful to start somewhere! -- I just have a few review comments below:

design/file-descriptors.md
sunfishcode: How about documenting them in the witx files, which are the sources from which the docs.md files are generated?

design/file-descriptors.md
sunfishcode: We don't have an answer for this yet :-}. POSIX often uses `-1` as an error code in functions that otherwise return a file descriptor. WASI APIs have a separate error return value, so they don't depend on special-case values like this. However, if we ever did return a file descriptor with this value, most C code would think it was an error.

design/file-descriptors.md
```diff
@@ -0,0 +1,14 @@
...
+
+# The Console
+When a program is initiated, `stdout`, `stdin`, and `stderr` (respectively `fd` ids `0`, `1`, and `2`) are opened. To print to these, simply call `fd_write` with the corresponding arguments. An simple "Hello World!" program written in `wat` is shown below:
```
programmerjake, line 5: Shouldn't the order of file descriptors be `stdin`, `stdout`, `stderr`? That's the way I've seen it on every other OS that I can think of.

design/file-descriptors.md
lachlansneff: Since handles is a more accurate term here since these can refer to pretty much any kind of resource (especially when imported types are part of wasm), should the terminology be completely replaced by "handle"?

design/file-descriptors.md
programmerjake: Note that Rust is in-progress defining `-1` as an invalid file descriptor, to the extent that assigning `-1` to a file descriptor type  would be undefined behavior: https://github.com/rust-lang/rust/pull/74699

This allows representing `Option<FileDesc>` with the layout of an `i32` and using the `-1` value to represent `None`.

design/file-descriptors.md
sunfishcode: Good catch -- yes, it should be `stdin`, `stdout`, `stderr`.

design/file-descriptors.md
sunfishcode: Eventually, handles should be represented in wasm by references, at which point we can start being more careful to use the term "file descriptors" to just mean integer indices into the handle table.

Concerning the name "file descriptor" applying to things that aren't files, I agree it's confusing, however "file descriptor" and "fd" are commonly used in POSIX and Unix-type platforms for things that aren't files, eg. in [`socket`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/socket.html). "Everything is a file", after all ;-). Given that we'll hopefully be saying "handle" most of the time, especially at the witx / documentation level, I think it makes sense to continue using the POSIX "file descriptor" and "fd" names to the extent that we're talking about integer indices, though it's admittedly subjective.

design/file-descriptors.md
sunfishcode: WASI has its own representation of `File` separate from the Unix definition, so that optimization as written doesn't apply to WASI. We could do the same optimization for WASI, but only if we decide to declare `-1` as invalid on WASI. On the surface it seems sensible; no typical host could support anywhere near a billion open files. And of course programs expecting POSIX semantics wouldn't be able to work with `-1`. My only concern here is in what happens as we generalize handles to support things other than files and high-level APIs other than POSIX and friends.

design/file-descriptors.md
SirJosh3917: Where are the witx files? Sorry if this is obvious and I'm just missing it.

design/file-descriptors.md
SirJosh3917: **TL;DR** nah

Personally, I think that less code is better from the perspective of the (Wasmtime/Wasi/Wasm/etc) library developer, and as such, there should not be more convenient methods. Here's why:

- **One way to do things.**
Consistency throughout the documentation is important, and by forcing people to go the iovec route you're ensuring things will be done correctly. Iovecs are the "correct" way to do things, because they discourage common performance pitfalls (i.e. string concatenation).
- **Smaller specification to conform to.**
By not having the overload, your interpreter does not grow in size. It stays small, and people who wish to implement the specification can do so easily and reasonably. Staying small in turn encourages widespread adoption. For example, the x86_64 standard is so huge and bloated that it's very difficult to create emulators for them (some are VirtualBox, Vmware, and QEMU) while a language such as `C` is adopted nearly universally at this point due to it's extreme simplicity (but its age also causes it to be complicated in its own rights). Admittedly, this comparison isn't very fair, but it's what I was able to think of on the top of my head.
On the contrary, for features that'd be too hard for an interpreter to get right, it should be implemented into the instruction set - such as vectorization. It just makes more sense for the tool generating the Wasm to generate code that can vectorize appropriately, and the cost of trying to write an interpreter to correctly take advantage of vectorization patterns would just lead to more bloat then letting the compiler do it.
- **Wasm has no need to optimize for the human.**
Just like how you'd never write assembly unless you're starving for those 0.27 nanoseconds, one should never have to write web assembly either. It's primarily generated by other compilers which target web assembly and whose generated outputs will get better as time progresses on. There's simply no need to optimize for the human here.
- **Encourages performant code.**
This is highly subjective, but what I imagine is that given an iovec is setup to use in the first place, it can most likely be reused in future operations. An overload for `read`/`write` would most likely (take this with a grain of salt because i only barely have an idea of what i'm talking about lol) have to setup its own iovec, copy the input into it, and then call `file_write` (or vice versa - setting up an iovec and marshalling the resulting bytes as whatever structure).
- **Having less allows more to be done.**
Instead of having an extra method or two and having to care for it and possibly any number of others that are introduced too as little helpers (documentation, ensuring correct code, deprecation of features, etc.), you can primarily focus the engineering effort that would be spent into maintaining that (as well as the other numerous projects that'd have to cope with the overload) into optimizing more pressing concerns - such as introducing a way to have no-copy writes (i.e. string into iovec being free, or some other datastructure which gets the job done in a similar fashion)

Apologies if any point is unclear or what have you, I feel like I rambled for far too long - hopefully that helps you make a decision. Of course, I'm not any designer at Wasm so I don't have a clear picture for the goals of the project or how it all works together as a whole, so read my comments with not one but two grains of salt.

Oh, and I added hello world example you mentioned.

design/file-descriptors.md
SirJosh3917: Fixed!

design/file-descriptors.md
SirJosh3917: I made it more explicitly clear that they're interchangeable with `handle`s. I'm not familiar with the in detail workings of Wasm/Wasi to choose on my own if I should include the nuanced discussion about what an `fd`/`handle` is at its core, so I chose to leave that complexity out for the time being.

design/file-descriptors.md
SirJosh3917: Gotcha!

design/file-descriptors.md
SirJosh3917: In that case, I'll opt to leave out a null file descriptor section until more detail are finalized.

design/file-descriptors.md
sunfishcode: Currently they live here: https://github.com/WebAssembly/WASI/tree/master/phases/ephemeral/witx

We'll likely be reorganizing things as WASI modularization moves forward, but this is where they are for now.

sunfishcode:
This looks like it's in good shape. I see it's still marked as a "Draft"; is there more to be done here?

SirJosh3917:
Ah yeah, I was going to move the documentation for specific code related to files to the WITX documentation, then link to it in the Files subcategory. Haven't gotten around to that yet, my bad - I'll try get it done relatively soon.

SirJosh3917:
Apologies for it having been a while since I've worked on this (so much for "relatively soon", eh?). I recall last stating that I needed to do this,

> Ah yeah, I was going to move the documentation for specific code related to files to the WITX documentation, then link to it in the `Files` header. Haven't gotten around to that yet, my bad - I'll try get it done relatively soon.

but can't quite seem to remember what that would entail (I'm unsure what I would link to in the `Files` section, where in the `witx` folder I would link to that section, and what code I would move over). Moreover, I assume there might be significant changes since I've last worked on this, so perhaps it'd be best to get another look at it.

programmerjake:
perhaps there should be some description added of how a file descriptor is closed, since just dropping an integer does nothing.

sunfishcode:
Thanks for picking this up again! Since this PR was started, a few developments have changed the landscape somewhat:

A major development since this PR was started has been the development of *handles* as an explicit feature of the interface-types proposal. The [canonical ABI](https://github.com/WebAssembly/interface-types/pull/132) PR includes an ABI which can represent handles as `i32` indices within modules, and includes resource lifetimes and closing handles. I expect we'll want to rebase WASI's concept of file descriptors on this new handle mechanism. That will let WASI focus on API functionality, and leave low-level resource management to lower levels of the stack that all APIs can use.

And, WASI is moving toward modularization. `path_open` is now defined in the [wasi-filesystem](https://github.com/WebAssembly/wasi-filesystem) proposal. I think that's the best place for documenting how to open and close files now.

Stdout, stdin, stderr are interesting. These won't be needed in all use cases, so I expect WASI will move toward a model where those will be defined by libc and tooling rather than by WASI itself. That said, we'll likely need to continue to support "commands" in their current form, which expect 0, 1, and 2 to be prepopulated. This isn't all figured out yet, but I'm starting to think of [wasi-classic-command](https://github.com/WebAssembly/wasi-classic-command) as the place where we'll collect documentation for how these kinds of commands work.

Does that make sense?

SirJosh3917:
Thanks for the information! I read more into what you mentioned, by reading the [Handle and Resources presentation](https://docs.google.com/presentation/d/1ikwS2Ps-KLXFofuS5VAs6Bn14q4LBEaxMjPfLj61UZE/edit#slide=id.p), the [Scoping and Layering presentation](https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8/edit#slide=id.p), and through the [canonical ABI](https://github.com/WebAssembly/interface-types/pull/132) proposal that was linked.

I decided to make this PR more geared towards resources and handles, as that seems like the replacement for file descriptors, and tried to write documentation for them based on my current understanding. I ended up dropping the mention to the console, but I should probably put that back in there if the commands proposal makes use of resources and handles.

sunfishcode:
As another update here, interface types has now evolved into a [component model](https://github.com/WebAssembly/component-model), which is now defining the concepts of handles, resources, and commands. Going forward, the component model will be the best place for documentation about what handles and resources are and how they work, with the WASI repo here focusing more on just the APIs that are built on top of them.

sunfishcode:
As I mentioned above, resources and handles are now defined by the [component model](https://github.com/WebAssembly/component-model).

PR #321, opened 2020-09-21
Add an agenda for the 09-24 meeting.

sunfishcode:


pchickey:
should we cancel the meeting for lack of an agenda?

PR #325, opened 2020-10-05
CI: use $GITHUB_PATH rather than add-path

pchickey:
In accordance with this advisory it's recommended we move to a
different scheme of setting env vars and updating PATH:

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


sunfishcode:
Looks good, thanks!

PR #330, opened 2020-10-08
Updated module-types links to module-linking links

MendyBerger:


jryans:
It would be great to merge this. I was about to send the same fix! 😄

tschneidereit:
Thank you for the PR, @MendyBerger, and for the prod, @jryans! Indeed, this is very straight-forward, so I'll go ahead and merge it.

PR #337, opened 2020-10-20
WASI Logo in next WASI meeting

syrusakbary:
Add proposal to cover WASI Logo in next WASI meeting

### DIFF
--- a/meetings/2020/WASI-10-22.md
+++ b/meetings/2020/WASI-10-22.md
@@ -26,5 +26,13 @@ Installation is required, see the calendar invite.
     1. Please help take notes.
     1. Thanks!
 1. Vote to move wasi-nn to stage 2; see [witx here](https://github.com/WebAssembly/wasi-nn/blob/master/phases/ephemeral/witx/wasi_ephemeral_nn.witx).
+1. Discuss about how to move forward with a [WASI Logo](https://github.com/WebAssembly/WASI/issues/3). Proposals:
+    1. We would never like to change the [current logo](https://github.com/WebAssembly/WASI/raw/master/WASI.png).
+    2. We would like to keep the [current logo](https://github.com/WebAssembly/WASI/raw/master/WASI.png) for now, and change it in the future when WASI is more mature.
+       TODO: Ask the risks of changing the logo short-term and see if there is any way to address them.
+       TODO: define what "mature" means (objectively) and how the process will be then
+    3. We would like to start researching on a new logo now.
+       TODO: define how the process will be
+    4. Other ideas?
 
 ## Meeting Notes


linclark:
> TODO: Ask the risks of changing the logo short-term and see if there is any way to address them.

I see an important risk that I want to raise. Unfortunately I can’t attend the meeting, so posting this here.

We would need to work with a lawyer to determine the IP issues around this. It’s possible that just using the same policy as for the WebAssembly logo works, but I know that various organizations have grown more concerned about IP in this space since then.

This is in no small part because your company, Wasmer, attempted to register “WebAssembly” and “Wasm” as trademarks. The [USPTO rejected the application](https://tsdr.uspto.gov/documentviewer?caseId=sn88703780&docId=OOA20200227112217#docIndex=3&page=1), but I don’t think we can assume that attempts to register a brandmark for a WASI logo would be similarly rejected. 

We need to consider the community’s long term best interests in any decision like this. If a single company were to acquire the trademark to WebAssembly, I believe that would be actively harmful to the community. I think the same applies for the brandmark rights to the WASI logo. 

Given the context, adopting a new logo would require a good amount of work on the part of the chairs—we need legal advice on what the IP policy should be, and the chairs would have to organize that. I think this is too much of a burden to put on them for something that is currently so limited in its impact.

syrusakbary:
> We would need to work with a lawyer to determine the IP issues around this

Happy to offer a copyleft version of any logo we propose. I believe that shall solve the IP concerns around it.
The goal of pushing the logo forward is not having a company to own it and sue other companies on its usage. The goal is to increase de adoption of WASI.

> This is in no small part because your company, Wasmer, attempted to register “WebAssembly” and “Wasm” as trademarks. The USPTO rejected the application, but I don’t think we can assume that attempts to register a brandmark for a WASI logo would be similarly rejected.

There is much more context than what is presented, and I'll do my best to shed some light on it.

The trademark requests were driven by our lawyers.
After getting contacted from someone in the community a few months ago, we reached them and decided to not pursue that path. This has nothing to do with the request for changes in the trademark request from USPTO (which was not a rejection), as we could have still pursued and owned the trademark after... (see Oracle owning the Javascript trademark).

We have decided this path even if it could be specially harmful for us in the case an organization like the Bytecode Alliance owns it in the future... why? Because of they have already veto-ed our intention to join based on personal and/or racial bias... trying to favor an ecosystem where Wasmer (the company) is out of the picture (sad, but true).

I think is important to note the following because some times is very easy to assume wrong intentions from players in the industry, specially if it fed into an argument that favors specific products (unfortunately it's been happening quite a lot for us):

The success of Wasmer as a company is highly tied to the success of WebAssembly as a technology. Therefore is in our best intentions to assure the long-term success of it, as long as accelerate it's adoption as much as we can on the short term.

------

We are the underdog here. I'd really appreciate if rather than fighting we could collaborate and make of WebAssembly and WASI a much more welcoming place, although that's no longer in my hands.

sbc100:
Sounds reasonable to at least have a discussion around this tomorrow.   Any idea how much time this might take @syrusakbary ?   I think most of the other agenda items we have for tomorrow will fairly short.  I will merge this now and if there are other short items I might put them first if that makes sense?

syrusakbary:
> Any idea how much time this might take @syrusakbary ?

I think probably max 10-15 min? Hopefully it will not take very long.

> I will merge this now and if there are other short items I might put them first if that makes sense?

Sounds good 👍

syrusakbary:
Here are the slides I presented today:
https://speakerdeck.com/syrusakbary/wasi-logo-proposal

I also followed up with a revisit of the current WASI icon in the original issue: https://github.com/WebAssembly/WASI/issues/3#issuecomment-714740192

Thanks everyone for your feedback during the presentation

PR #339, opened 2020-10-21
Add an agenda item for a presentation on character encodings.

sunfishcode:
There are several discussions going about character encodings and
strings across different forums. This presentation will summarize
the current conversations and discuss the high-level pros and cons
of various options being discussed.


tschneidereit:
Is there any chance this item can be moved up above the one on the logo? I unfortunately will only be able to join after half an hour, and would like to be present for the other topic.

sunfishcode:
Sure; I've now moved it up.

sbc100 (approved):

PR #342, opened 2020-10-22
add action items

sbc100:


sbc100:
Where there action items from the last meeting Dan?   (I was not in attendance).

Issue #357, opened 2020-11-25
witx: What is `supertypes` in `HandleDatatype`?


yagehu:
[`HandleDatatype`](https://github.com/WebAssembly/WASI/blob/52cb4fea61b20cf87e15653e0e3df28fe8b432e6/tools/witx/src/ast.rs#L298-L300) is defined as:

```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct HandleDatatype {
    pub supertypes: Vec<TypeRef>,
}
```

What is `supertypes`? The only use of `handle` in snapshot preview is:

```
(typename $fd (handle))
```

pchickey:
At one point I had the poor idea to allow handles to be a subtyping hierarchy but fortunately we do not actually use this feature anywhere. It turns out this isnt compatible with using wasm extern refs as handles. Ignore the field, and I will remove it from the language when I get a chance.

yagehu:
Thanks. Answered my question.

PR #358, opened 2020-11-30
witx: remove supertypes from the handle type

pchickey:
the idea here was to enable handles to have a subtyping hierarchy,
but this isnt compatible with wasm extern refs and we never used it
anywhere, so its safe to delete all this code.

See also #357

### DIFF
--- a/tools/witx/src/ast.rs
+++ b/tools/witx/src/ast.rs
@@ -295,9 +295,7 @@ pub struct UnionVariant {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct HandleDatatype {
-    pub supertypes: Vec<TypeRef>,
-}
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):

sunfishcode:
The subtype relationships are meant to anticipate [subtyping in type imports](https://github.com/WebAssembly/proposal-type-imports/blob/master/proposals/type-imports/Overview.md#subtyping).

Type imports are better than externref for WASI because externrefs aren't virtualizable, since they can only be created by the host, so it'd be good to keep the subtype relationships in anticipation of using them with type imports.

pchickey:
oh! as usual I got it backwards. I think i'd like to keep this code out for YAGNI reasons. If we need to add it back in, its not too difficult to do so, either by revert or manually.

Issue #360, opened 2020-12-02
Repository organization


sunfishcode:
Currently the main directory organization in the WASI repository looks like this:

`phases/`
   - `ephemeral/`
   - `snapshot/`
   - `old/`
      - `snapshot_0/`

I've recently talked with several folks about how to better organize proposal repos, and we came up with several ideas which seem to fit well together which I'd like to discuss:

 - Switch to having stable URLs for snapshots rather than maintaining a `snapshot/` pointing at the current snapshot and `old/` add old snapshots. We can have a link in a README file pointing to the latest revision if that would help, but being able to refer to a specific snapshot URL would be useful. This was suggested in https://github.com/WebAssembly/WASI/issues/261, and I now think it makes sense to do in combination with the other ideas here.
 - Replace numbered snapshots ("0", "1", etc.) with date-based snapshots (eg. "2020-12" or so)
 - Replace the contents of the `ephemeral/` directory with a proposal template. The way this will work is:
    - Development will move into the proposal repos, such as wasi-filesystem.
    - When we create a new proposal repo, it'll get a copy of the template, and the people working on the proposal can use that if they wish, but they can also choose to organize development however they see fit.
    - We can also then rename this to `proposal-template/` and move it out of `phases/`.
 - Add a new `standard` directory, which in the main WASI repo here will be empty until some proposal reaches stage 5, but which in the proposal repo forks will contain the most recent version of the proposal.
 - Add a script, similar to `update-testsuite.sh` in [the wasm testsuite repo](https://github.com/WebAssembly/testsuite), which copies the `standard/` contents from the proposal repos into a date-based `snapshots/` subdirectory.

So the new structure would be:
 - `phases/` goes away
 - add `proposal-template/` at the top level contains a template for new proposals
 - `snapshot/`
    - date-based snapshot directories (eg. 2020-12)
 - `standard/`

This will, at last, be part of a clear path forward for releasing a new snapshot using the proposal repos. Thoughts?

syrusakbary:
I'm not super fan of date-named versions since date doesn't convey meaning.

However if date-named versions are the way to go I'd prefer using a convention such as `Dec2018` rather than `2020-12` (following a similar nomenclature than the GraphQL spec, for example).

sbc100:
One good thing about `2020-12` it sorts.   I prefer it over `Dec2018` but maybe I'm biased because I'm very used to ubuntu's naming scheme of `16.4` which is very similar to `2016-12`.

[...]

PR #364, opened 2020-12-10
Reorganize the directory structure to better support proposals.

sunfishcode:
Following the organization sketched out here:

https://github.com/WebAssembly/WASI/issues/360

This PR adds several new directories and starts a work-in-progress
make-snapshot.sh script for creating snapshots.

The `phases` directory is not yet removed; that can happen once
we've transitioned everything to the new organization.


linclark:
This looks great!

One question I had: Do we want to check in the `repos` directory? It seems like we could just allow that to be a transient directory that is created at the start of the script to collect the repos to work on and then gets blown away at the end.

sunfishcode:
Thanks! I've now removed `repos`. I'll merge this now, and we can continue to iterate from here.

PR #371, opened 2021-01-06
Add the `(@witx noreturn)` annotation to proc_exit

pchickey:
The noreturn annotation was invented solely for the sake of proc_exit,
but we never actually used it in the spec documents.

This annotation sets the `noreturn` bit in the `InterfaceFunc` struct.
Downstream tools are free to do with that bit what they will.


sbc100 (approved):

sunfishcode:
LGTM too; looks like the CI just wants a few test cleanups as well.

PR #377, opened 2021-01-20
Add mechanism for displaying the WAT signatures of WASI modules

abrown:
@pchickey, here are the beginnings of an idea for displaying signatures of WITX interfaces. What I need is some way to easily determine what WASI modules will look like after the WITX is translated away. I felt like WAT was a good format for this, though ideally I would want to generate the same type of information in other languages (e.g. as C headers or AssemblyScript declarations).

Currently, when I run `cargo run --example witx -- wat-signature ../../phases/ephemeral/witx/wasi_ephemeral_nn.witx`, this PR outputs something like the following:
```
(module
 (import "wasi_ephemeral_nn" "load" (func (param I32 I32 I32 I32 I32) (result I32)))
 (import "wasi_ephemeral_nn" "init_execution_context" (func (param I32 I32) (result I32)))
 ...
)
```

I am seeking feedback on whether this type of helper command is useful to others before I invest more time to fix it up (e.g. use a proper WAT library, add comments, fix type casing, etc.).

### DIFF
--- a/tools/witx/examples/witx.rs
+++ b/tools/witx/examples/witx.rs
@@ -42,6 +42,12 @@ enum Command {
     },
     /// Update documentation in WASI repository to reflect witx specs
     RepoDocs,
+    /// Output the expected WAT signature of a witx module
+    WatSignature {
+        /// Path to root of witx document
+        #[structopt(number_of_values = 1, value_name = "INPUT", parse(from_os_str))]
+        input: Vec<PathBuf>,
+    },
     /// Examine differences between interfaces
     Polyfill {
         /// Path to root of witx document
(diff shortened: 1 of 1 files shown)


pchickey (commented):
This looks useful and I'm glad that its a very compact and easy thing to implement so far. I don't have a strong opinion about whether it should use a library to help render it - the current witx sexpr-rendering code does not, and probably should. Thank you for doing this!

abrown:
> the current witx sexpr-rendering code does not, and probably should

Is this the ideal way to proceed? E.g. I see that [SExpr](https://docs.rs/witx/0.8.8/witx/enum.SExpr.html) implements `Display`--should I build up an sexpr for each signature and print those?

pchickey:
That is the "good enough" solution I got working, but its still a pretty unpleasant output for a human to read because it doesn't try to put line breaks or whitespace in sensible places. 

@alexcrichton are you aware of any s-expr pretty printing libraries we could use?

alexcrichton:
Nah unfortunately not that I'm aware of :(

Crates like `wasmprinter` just hand-roll their own management of indentation levels and such.

abrown:
As discussed in today's WASI meeting (cc: @sunfishcode, @lukewagner), if/once https://github.com/WebAssembly/interface-types/pull/132 is merged then this PR should be replaced by some tooling to do some form of "WITX-> WIT" conversion (i.e. generating an official core-wasm signature for the high-level WITX).

sunfishcode:
Since this PR was filed, a lot has changed. Using [wasm-tools](https://github.com/bytecodealliance/wasm-tools/), the signature of a component can now be printed using `wasm-tools component wit`.

PR #388, opened 2021-01-31
Fixed typo in random_get documentation

guyroyse:


jedisct1:
Good catch!

The markdown files are automatically generated from the `witx` files. So, the change should be made there.

guyroyse:
The typo is in two of the witx files that I can see, the ephemeral and the snapshot. I'll send an updated PR.

sunfishcode:
Thanks for reporting this! It's admittedly a little confusing because we're in the middle of a transition away from having everything in this repo towards having proposal repos, with the random APIs in WebAssembly/wasi-random. I've now created WebAssembly/wasi-random#4 to port this fix to the wasi-random repository (and added you in the Co-authored-by).

guyroyse:
Thanks! I shall close this PR then.

PR #391, opened 2021-02-08
Refactor the AST of types in the `witx` crate

alexcrichton:
This series of commits is intended to start a refactoring of the `witx` crate's AST to get, yet again, closer to interface types. With a [recently rereshed explainer](https://github.com/WebAssembly/interface-types/blob/master/proposals/interface-types/Explainer.md#interface-types) interface types can almost entirely encompass what WASI uses right now with a smaller set of types and some more shorthands. The purpose of this PR was to simplify all types to just those specified in the interface types specification right now, namely:

* `Array` is renamed to `List`
* `Struct` is renamed to `Record`
* `int` typedefs are refactored into simple builtin types and a separatetly-stored list of constants
* `Flags` as-is are redefined as integers with associated constants for each value. Note that this is different from the proposed elaboration of `flags` to a struct-of-bools. While we may want to go in this direction eventually I opted to do the simple thing to keep ABI stability for now.
* `string` is recast as `list[char]`
* `Union` is entirely defined in terms of `Variant`
* Types like `Char8` and `Usize` are now hints on `U8` and `U32`

The general goal is to get the type-level AST to exactly match interface types. The eventual intention is to continue refactoring internals and exposing more functionality around "what is the wasm signature for this function" along with "how do I convert a language value to this wasm signature?" or "how do I convert the wasm signature to a language value?".

This is not intended to have any ABI-breaking changes at this time. It's a large breaking change for tooling, though, since lots got renamed. I've been updating the [wasi crate](https://github.com/bytecodealliance/wasi) as part of this work to ensure that the same ABI is generated as before.

I've added a number  of locations in definitions of flags/enums as well that have specific `@witx` annotations. These are mostly around tag sizes for now, to specifically say that the size of the tag on a `Variant` is required to be a particular value. Eventually interface types will abstract the need for this to be specified at all, and more closely in the future I hope to start proposing a stable ABI mapping from interface types used by witx to wasm signatures, where the stable mapping would mean we could eschew manually sizing tags as well. 

I hope to follow-up with subsequent PRs to this as well. Namely I would like to explore updating the internals to centralize more details about wasm-to-language and language-to-wasm conversions in this repository. I'm not 100% sure how this will look yet but the intention is that it should be easy to use the `witx` crate to be used for any language binding generator to convert signatures back and forth.


tools/witx/src/ast.rs
```diff
@@ -209,12 +211,27 @@ impl Type {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum BuiltinType {
-    String,
-    Char8,
-    USize,
-    U8,
+    Char,
```
sunfishcode, line 214: Could you add a comment here mentioning that this `char` is a Unicode scalar value?

tools/witx/src/ast.rs
```diff
@@ -209,12 +211,27 @@ impl Type {
...
+    U8 {
+        /// Indicates whether this type is intended to represent the `char`
+        /// type in the C language. The C `char` type is often unsigned, but
+        /// it's language-specific. At an interface-types level this is an
+        /// unsigned byte but binding generators may wish to bind this as the
+        /// language-specific representation for a C character instead.
+        lang_c_char: bool,
```
sunfishcode, line 221: The `char8` type in witx is meant to roughly correspond to [`char8_t`](https://en.cppreference.com/w/cpp/language/types#Character_types). Besides signedness, `char8` indicates an intent to point to Unicode string data.

tools/witx/src/docs/ast.rs
```diff
@@ -356,12 +288,10 @@ impl ToMarkdown for InterfaceFuncParam {
...
         match self {
-            BuiltinType::String => "string",
-            BuiltinType::Char8 => "char8",
-            BuiltinType::USize => "usize",
-            BuiltinType::U8 => "u8",
+            BuiltinType::Char => "char",
+            BuiltinType::U8 { .. } => "u8",
             BuiltinType::U16 => "u16",
-            BuiltinType::U32 => "u32",
+            BuiltinType::U32 { .. } => "u32",
```
sunfishcode, line 294: Would it make sense to check `lang_ptr_size` here and map to `"usize"` as appropriate?

phases/ephemeral/witx/typenames.witx
```diff
@@ -614,24 +612,24 @@
...
 ;;; The contents of a `subscription`.
 (typename $subscription_u
-  (union $eventtype
-    (field $clock $subscription_clock)
-    (field $fd_read $subscription_fd_readwrite)
-    (field $fd_write $subscription_fd_readwrite)
+  (union (@witx tag $eventtype)
+    $subscription_clock
+    $subscription_fd_readwrite
+    $subscription_fd_readwrite
```
sunfishcode, line 626: It's not immediately clear why some `union`s such as `event_u` are migrated to `variant` and others, such as `subscription_u` here, are left as `union`.

sunfishcode (commented):
Overall looks great!

tools/witx/src/ast.rs
alexcrichton: I added a comment along these lines, but did you think we needed other changes? I'm hoping that we can eliminate the `char8` type as we move closer to pure-interface-types wasi.

phases/ephemeral/witx/typenames.witx
alexcrichton: The desugaring is a bit tricky with the current ABI, but I generally tried to keep things as `union` where possible. By using `@witx tag` the `union`'s types must all mattch up exactly with the variants of the `tag` (which should be an `enum`).

One place this was tripped up though was the `empty` case parsing for unions. That was where one entry of the `union` didn't have any payload. This didn't have any translation to this updated `union` syntax where it must be a set of types that are union'd, so anything using `empty` was switch to the explicit `variant`.

In general it doesn't matter too too much except from a readability perspective since this all desugars to the same thing in the AST, which is to say a tagged `variant`.

sunfishcode: Is `variant` a superset of the functionality of `union`? If so, I think it's ok to migrate to `variant` over time. To help readers, I do want to aim for a single syntax, but we can finish the migration later.

tools/witx/src/ast.rs
sunfishcode: Agreed; we should ideally use interface-types' `string` type for strings wherever possible.

sunfishcode (approved):

phases/ephemeral/witx/typenames.witx
alexcrichton: It is, yeah, mostly because `variant` is the actual thing in the AST whereas `union` is just syntax sugar as [seen here ](https://github.com/WebAssembly/interface-types/blob/master/proposals/interface-types/Explainer.md#interface-types).

This will hopefully become less important over time as `@witx tag` is migrated away from perhaps? Otherwise though there could perhaps be more syntax sugar for a `union` where one of the cases is empty.

PR #392, opened 2021-02-10
Add a new style of test suite for the `witx` tooling similar to `*.wast`

alexcrichton:
These commits add a new `*.witxt` style of test suite to be similar to the `*.wast` style of test suites. (supposed to be shorthand for "witx test")

Most existing tests have been converted to the `*.witxt` style of test as an example. There's a few goals from this translation:

* In general adding a few lines to a `*.witxt` file is quite easy and low-overhead. The intention is to make the barrier-to-entry on adding test extremely low.
* Avoid API churn. By having a separate textual format for tests it helps us reduce churn on the tests themselves if the APIs of the `witx` crate change.
* As the functionality of `witx` grows and expands this is intended to make it easier to add new classes of test suites. We'll in theory already have a lot of witx documents in the test suite already and are used to writing inline witx documents, so adding future functionality should be much easier to test by either reusing existing test directives or adding a small handful of new directives.

For example one thing I'd like to add stricter tests for on https://github.com/WebAssembly/WASI/pull/391 is the elaboration of parsed types to the variant/record types. By using `assert_representation` I could reuse those definitions and existing type checking to assert this by adding a new text file here. Additionally I'd like to start codifying more of the translation of a witx signature to the expected ABI, and I believe that will be much easier with a test suite like this.

Another thought on this style of test suite is that as `*.witx` usage grows beyond simply WASI we'll want the ability to test more and more pieces of functionality that aren't necessarily used by WASI itself, which means it'll get more important to have strict and exhaustive tests outside of just the `*.witx` files in this repository for WASI itself.


pchickey:
I love this! I think there is a typo in your second para...

pchickey (approved):
Lets release this as witx 0.9?

alexcrichton:
Ok with https://github.com/WebAssembly/WASI/pull/391 merged I've rebased this on master to get all the tests passing again with adjusted syntax. This actually helped uncover a few bugs here and there!

Unless there's an anything urgent I'd prefer to not have a new release just yet. My hope is to next start implementing conveniences for code generators (e.g. witx-bindgen in Rust, wiggle in Wasmtime, the C header generator in wasi-libc, ...) to largely not have to worry about these changes. This is still a work in progress but it'd be nice if they didn't have to update twice or land in the middle by accident.

alexcrichton:
I think this should be good to merge now though!

pchickey (approved):

PR #393, opened 2021-02-11
Fix proposal links

abrown:


abrown:
@sunfishcode, can you take a look?

sunfishcode:
Thanks!

PR #397, opened 2021-02-23
Initial checkin of test scripts, copied from wasi-sdk.

sunfishcode:
This just copies in the test scripts from https://github.com/WebAssembly/wasi-sdk/tree/main/tests. We can evolve them from this point on.


sunfishcode:
This is a first draft at a concrete setup for the testing process described in #373. I copied in the same test script that #373 describes and set up a basic directory structure, which hopefully makes it clear where things will go in the repository.

caspervonb:
Got some things I want to change here but I'll follow up with that.

LGTM, good to get the ball finally rolling on this.

sbc100:
Is the idea that wasi-sdk will then get its test runner from the WASI submodule (and we can then delete its copy).?

sunfishcode:
The wasi-sdk version of these scripts has support for running the tests with different optimization levels, and also compile-only tests, but I think what we can do is build those on top of the scripts here.

sunfishcode:
I'm going to take the liberty of merging this, so we can proceed to iterate on top of it. If there are further comments, please file issues in the issue tracker!

PR #398, opened 2021-02-23
make witx cli a new crate, refactor how we validate repo contents

pchickey:
These changes are designed to make the witx tooling more useful to proposal forks of this repository.

* Rather than keep the cli-oriented functionality of the `witx` crate in an example, make a new crate `witx-cli` that creates a `witx` executable.
* Add a `--check` flag to the `witx docs` subcommand that checks if docs in the filesystem match those calculated from the witx files.
* Rewrite the `wasi-docs` test to invoke `witx docs --check`. This test now can live in .github/workflows/main.yml rather than in the rust sources.
* Testing of the witx tooling and validating witx specs & docs can now happen in parallel in CI.
* Speaking of things that don't belong in the rust sources: paths to witx files don't belong in there, so delete the `witx::phases` module. Really they never belonged in there, but I didn't see the harm before we started the repo re-organization in #360.


tools/repo_docs.sh
```diff
@@ -0,0 +1,15 @@
+cd $(dirname $(realpath $0))/witx
+cargo run -p witx-cli -- docs ../../phases/snapshot/witx/wasi_snapshot_preview1.witx --output ../../phases/snapshot/docs.md
```
alexcrichton, line 2: Could this be called from CI to avoid duplicating it?

alexcrichton (approved):

tools/repo_docs.sh
pchickey: Good point. I'll add an argument to pass in --check on CI

PR #399, opened 2021-02-24
Document `=` and `\0` use in args/environ docs.

MaulingMonkey:
Documents the answers I recieved in https://github.com/WebAssembly/WASI/issues/396 .  Hopefully I'm doing this right!


pchickey (approved):
Thank you! This is very helpful.

Issue #322, opened 2020-10-02
UTF-8 validation


sunfishcode:
UTF-8 is a very popular string encoding, for example, it's the encoding used by [over 95% of all Web content](https://en.wikipedia.org/wiki/UTF-8#Adoption). It's not uncommon for applications to need to do UTF-8 validation on their own strings, and since all WebAssembly VMs have UTF-8 validation logic built in as required by the spec, we should define a WASI API to let applications call into the VM's UTF-8 validation logic rather than having to bundle their own.

I'm picturing an API which takes a byte slice as input and returns a boolean value indicating whether it's valid or not. This is the minimum that WebAssembly engines themselves are required to have, and would be enough for eg. the use case of implementing a UTF-8 validity check for a a WASI API implemented in wasm.

More elaborate APIs are possible, such as validation which returns the position where an error occurred, and possibly information about the error, but I think it makes sense to start with something simple. I won't have time to make an official proposal myself for a while, but I wanted to file this issue to see what others think!

sunfishcode:
Thinking about this more, I think it may make sense to have a return value that carries more information.

Taking inspriation from Rust's [`str::from_utf8`](https://doc.rust-lang.org/stable/std/str/fn.from_utf8.html) and [`Utf8Error`](https://doc.rust-lang.org/stable/std/str/struct.Utf8Error.html), I'm picturing an API like this:

```wasm
(typename $utf8_error
    (record
        (field $valid_up_to $size)
        (field $incomplete bool)
        (field $error_len $size)
    )
)

...

(@interface func (export "validate")
    (param $bytes (list u8))
    (result $error (expected (error $utf8_error)))
)

```

jedisct1:
Hi,

I'm a little bit confused by this.

Why would a language's standard library call this instead of using the UTF-8 validation function they already have, and use for all other targets?

Pauan:
@jedisct1 Because of the way that the web works, file size is very important, and so relying on the browser's functionality means smaller file sizes.

PR #420, opened 2021-03-29
Add an `option` shorthand to witx.

sunfishcode:
I was mocking up a witx description for #322 and had a use for an
`option` type. Of course, I could also just emit a variant manually,
or reuse `expected` by omitting the error type, but `option` better
conveys the sense that the `none` case doesn't necessarily represent
an error.

### DIFF
--- a/tools/witx/src/abi.rs
+++ b/tools/witx/src/abi.rs
@@ -274,6 +274,15 @@ def_instruction! {
         /// WASI and intentionally differs from the type-level grammar of
         /// interface types results.
         ResultLift : [1] => [1],
+        /// Pops a native wasm `i32` from the stack, as well as one block
+        /// internally from the code generator.
+        ///
+        /// If the value is 0 then the "some" block value should be used.
+        ///
+        /// Note that this is a special instruction matching the current ABI of
+        /// WASI and intentionally differs from the type-level grammar of
+        /// interface types results.
+        OptionLift : [1] => [1],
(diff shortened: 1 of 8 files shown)


jedisct1:
That would be a nice addition!

In `wasi-crypto`, a custom `Option` type is also used everywhere, which leads to non-idiomatic generated code in languages supporting optional types.

alexcrichton:
The code here looks fine to me, but I would hesitate because I'm not sure whether we want to continue to enhance the current ABI of WASI since it makes it more complicated to support in a future with support for a new ABI. I've been working on an [`abi-next`](https://github.com/alexcrichton/wasi/tree/abi-next) branch locally which [also has support for `option`](https://github.com/alexcrichton/WASI/blob/2beb50f9e114762a550a80594cfe453a3a66a54c/tools/witx/tests/witxt/shorthand.witxt#L61-L65), and I definitely agree it would be nice to support!

Do you think it would be better for me to try to land by "abi next" work and get `option` supported only there? Or do you think that it would be best to stick with this extension to the current ABI?

sunfishcode:
Ah, cool. I'm ok waiting for the new ABI; this was just for some prototyping I was doing.

alexcrichton:
Ok! I'll work on cleaning that up for a PR this week.

alexcrichton:
Ok I posted https://github.com/WebAssembly/WASI/pull/422 which has `option` in it as well

PR #421, opened 2021-03-30
Enable more than one `handle` type in `*.witx`

alexcrichton:
This commit reworks how handle types work at the representation level in
`*.witx` file to follow the expected design of resources in the
interface types specification. Previously handle types were simply
`(handle)`, which meant that there could only be one actual handle type
in the world (structurally at least). After this commit, however, there
can be multiple handle types.

First abstract types must be introduced as a `resource`, for example:

    (resource $fd)

This declares that the module exports a type named `$fd` and it's
abstract in that the representation is not known. At the interface layer
you can't pass an `$fd` directly because it's representation is not
known. To do that, however, you can do:

    (param $x (handle $fd))

The `handle` type now refers to a particular `resource` that it refers
to. Values of type `handle T` can exist and are what's passed at the
boundaries.

This is all largely just an internal structuring concern at this point.
This has no ramifications for WASI which still has an `$fd` type for
functions that is represented with an `i32`. This is largely a
forward-looking change to allow multiple types of resources defined by
different modules and all used by one another.

This commit also updates `use` syntax where `use` will pull from either
the type or the resource namespace. Furthermore a new `(use ($foo as
$bar) ...)` syntax was added to locally renamed something within a module.

### DIFF
--- a/phases/ephemeral/witx/typenames.witx
+++ b/phases/ephemeral/witx/typenames.witx
@@ -281,8 +281,9 @@
   )
 )
 
+(resource $fd)
 ;;; A file descriptor handle.
-(typename $fd (handle))
+(typename $fd (handle $fd))
 
 ;;; A region of memory for scatter/gather reads.
 (typename $iovec
--- a/phases/old/snapshot_0/witx/typenames.witx
+++ b/phases/old/snapshot_0/witx/typenames.witx
@@ -273,8 +273,9 @@
   )
 )
 
+(resource $fd)
 ;;; A file descriptor handle.
-(typename $fd (handle))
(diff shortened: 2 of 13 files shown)


sunfishcode (approved):
Looks great!

It appears that handles aren't yet displayed in the docs.md; would it be a lot of work to add that? In any case though, I don't think it's urgent, because the doc comments on the handle types are where the main documentation is.

pchickey (approved):

alexcrichton:
Ah that's true, although it wouldn't be too hard to add documentation for `resource` types. If/when the documentation moves there I'm happy to update the docs generator to do so!

PR #433, opened 2021-06-15
witx: Update wast dependency

olivierlemasle:
witx: Update `wast` dependency

### DIFF
--- a/tools/witx/Cargo.toml
+++ b/tools/witx/Cargo.toml
@@ -17,7 +17,7 @@ crate-type=["rlib"]
 anyhow = "1"
 log = "0.4"
 thiserror = "1.0"
-wast = { version = "33.0.0", default-features = false }
+wast = { version = "35.0.2", default-features = false }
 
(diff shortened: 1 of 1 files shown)


pchickey:
@olivierlemasle Just curious, are you using the current `main` version of witx? Alex and I have been talking how the changes since 0.9.0 are sort of an evolutionary dead end, with some really major renovations coming soon to support interface types. We were even thinking of reverting the changes to `use` to reduce churn. But, if we have other users who care, we can tag the current state of things as 0.10.

pchickey (approved):

olivierlemasle:
@pchickey Actually no, I'm not using current `main`.
I'm packaging `witx` 0.9.0 for Fedora in order to package `wasmtime` (which is using version witx 0.9.0). I had to patch `witx` to use the Fedora-packaged version of `wast` (= the latest), and I pushed this patch upstream, in `main`, but I do not depend on current `main`.

Thanks for asking!

PR #437, opened 2021-06-22
Bump to 0.9.1

alexcrichton:
I'll publish this to crates.io after merging, and I also just
double-checked to ensure we don't have any other accidental API breakage
leaking in.


pchickey (approved):
Needs witx-cli as well

pchickey (approved):

Issue #450, opened 2021-10-20
Make release process more modular


linclark:
As discussed in the June 17 meeting (with an update on October 21), we have been working to make it easier to propose WASI APIs. This includes the previously discussed work on a more ergonomic syntax for defining APIs, a proposal template, and a simplified repo structure, which are all close to completion.

In addition to those changes, we need to look at the release process. The plan for WASI has evolved since its inception, but the way we thought of the release process didn’t change to reflect that.

## Problems with the current release process

WASI is a much more modular set of APIs than are traditionally part of a system interface. But our process is to group all of these APIs into a single “snapshot” for release.

This causes the following problems.

1. A snapshot has a single integer release number. With all proposal updates being grouped under this single version, there’s no sensible way to use semantic versioning to communicate about API changes. 
1. Most hosts will only support a subset of APIs, yet we bundle all the APIs into a single file. This means that consumers need to add a processing step to remove unnecessary APIs.
1. The snapshots are in the main WASI repo, even though the work happens across different repos. Creating a single artifact in the WASI repo creates confusion about where issues should be filed.

## Proposed new release process

Let’s look at another possible way of doing things. Instead of releasing a monolithic snapshot and versioning that snapshot, each API would cut its own releases. 

These releases would still go through the WASI standardization process, which means that we would vote before major and minor releases. These releases would then be listed in an additional column in Proposals.md.

This release process would maintain the rigor of the standardization process, but would do so in a way that reflects the modularity of the APIs and addresses the issues above by:

1. Making it possible for APIs to follow semantic versioning
1. Allowing hosts to easily specify exactly which APIs that they want to support before feeding them through tooling
1. Providing a single source of truth (the proposal repo) for all APIs

I will be bringing this proposal up for discussion in the October 21 meeting, and if no major concerns come up during the conversation, plan to propose a vote for the subsequent meeting.

syrusakbary:
> Regarding semantic versioning in particular, there was quite a lot of discussion around this early on I think we decided it didn't make sense to use this. 

Here's the previous discussion: #360

Issue #382, opened 2021-01-26
Move all WASI modules to separate repositories


linclark:
The ideal end goal of WASI modularization is for each of the modules contained in this repo to have its own repo.

Once we have the syntax in place for depending on types from other directories (#378), we need to move the witx files for the different modules to their own repositories.

sunfishcode:
We have now migrated all the proposals to their own repos, such as [wasi-filesystem](https://github.com/WebAssembly/wasi-filesystem) and [wasi-sockets](https://github.com/WebAssembly/wasi-sockets).

PR #451, opened 2021-10-20
Simplify main WASI repo

linclark:
As discussed before in meetings, we've received feedback that the current structure of the WASI repo is confusing. 

This PR significantly reduces the complexity of the repo structure. It does this by moving parts of the repo that will be obsoleted to a `legacy` directory, and editing and consolidating the rest.

It also introduces the idea of a [proposal template](https://github.com/linclark/wasi-proposal-template), which I've created a first draft of.

This version of the restructure depends on #450.


sbc100:
Isn't wasi-libc still based on `tools/witx` and the files in `phases/snapshot`.   I guess `tools/witx` is moving somewhere?    And the witx files it processes are going to come from other repos?   I think it might be good to show `wasi-libc` using whatever the new things is before we delete the old thing here.

lygstate:
> > I guess tools/witx is moving somewhere?
> 
> Alex has been working on a tool to replace `tools/witx`. This is the tool that supports the new syntax which he's been demonstrating in meetings over the past couple of months. The original `tools/witx` will move to the legacy directory.

Where is the tool, when the tool to be merged?
> 
> > And the witx files it processes are going to come from other repos?
> 
> Correct. This is a [change that has been in the works](https://github.com/WebAssembly/WASI/issues/382) since late last year with the push towards modularization.
> 
> > I think it might be good to show wasi-libc using whatever the new things is before we delete the old thing here.
> 
> This makes sense to me.

linclark:
> Where is the tool, when the tool to be merged?

With this simplification, the tooling would live outside of this repo, in the [wasi-tools repo](https://github.com/WebAssembly/wasi-tools) (which Dan just created). Once that's done, the main WASI repo would focus on coordinating work at a higher level and helping people navigate to the various APIs that are a part of WASI.

Alex is prioritizing getting the replacement tooling up in that repo, so I expect it to be available there soon.

linclark:
We won't be able to make a clean break with `witx` yet, so I updated this PR to leave the `phases` and `tools` directories in place. They will move to legacy once `wit` supports all APIs.

I'd like to merge this soon so that we're pointing new contributors in the right direction, so I'd ask that any word-smithing or nits be tackled in follow up PRs and that we focus comments on this PR on more substantial, structural concerns.

PR #454, opened 2021-11-23
Fix broken link in the proposals list

bnjbvr:


linclark:
Thanks for catching that!

PR #457, opened 2021-12-01
Update docs.md

ailisp:
According to man
```
       int stat(const char *pathname, struct stat *statbuf);
       int fstat(int fd, struct stat *statbuf);
       int lstat(const char *pathname, struct stat *statbuf);
...
       int fstatat(int dirfd, const char *pathname, struct stat *statbuf,
                   int flags);
```
`path_filestate_get` should be closer to `fstatat` instead of `stat`


sunfishcode (approved):
Looks right. Good catch!

Issue #370, opened 2020-12-28
Networking API design


badeend:
I've had some thoughts, and here's my braindump. Feedback is welcome.

## 1. Which use cases should the WASI networking APIs account for?

The official WebAssembly website mentions lots of use cases: https://webassembly.org/docs/use-cases/. I've kept the cases that seem relevant and added some of my own:
- Standalone desktop & server Wasm apps.
- IoT devices running microcontrollers without a full blown OS.
- Web browsers.
- container-based serverless providers like AWS, GCP and Azure.
- JavaScript/Wasm-based serverless providers like Fastly's Compute@Edge, Cloudflare Workers and Akamai EdgeWorkers. These look to be more locked down than their container-based counter parts.
- [HTTP proxies](https://github.com/proxy-wasm/spec).
- plugin/extension-style embeddings into existing applications.
- Audio / video streaming.
- Gaming
- Peer-to-peer applications (games, collaborative editing, decentralized and centralized).
- Image recognition.
- Platform simulation / emulation (ARC, DOSBox, QEMU, MAME, …).
- Language interpreters and virtual machines.
- POSIX user-space environment, allowing porting of existing POSIX applications.
- Remote desktop.
- VPN.
- Local web server.
- Fat client for enterprise applications (e.g. databases).
- Server-side compute of untrusted code.
- Symmetric computations across multiple nodes

## 2. Which protocols should be focused on first?

For reference, I've compiled a non-exhaustive list of commonly used protocols:

UDP, TCP, DNS, HTTP1/2/3, QUIC, TLS, SSH, SFTP, FTP(S), SMTP, POP3, IMAP, WebRTC, WebSockets, gRPC, UNIX sockets.

#### Outside the browser

Many runtimes (Java, .NET, ...) have built their own HTTP stack on top of the OS provided socket API's. Even if WASI was to provide an HTTP interface, it seems unlikely existing codebases would switch to it anytime soon.

#### Inside the browser

Browsers already have an HTTP API: `fetch()`. All WASI can provide is a wrapper for that function.

Also, all WASI's current proposals are, as far as I can see, blocking synchronous functions. And browsers won't swallow that. Until there is an general answer for asynchronicity in Wasm/WASI, I propose to leave this use case out of scope for the networking proposal.

### My thoughts

TCP and UDP are the lingua franca of the web with pretty much every other protocol in existence built on top of that. It seems logical to start with these, although I'd be happy to be proven wrong.

Pros:
[...]

PR #461, opened 2021-12-28
Add link to networking proposal.

badeend:
In response to issue #370

### DIFF
--- a/docs/Proposals.md
+++ b/docs/Proposals.md
@@ -49,6 +49,7 @@ Proposals follow [this process document](https://github.com/WebAssembly/WASI/blo
 | ------------------------------------------------------------------------------ | -------------------------------------- |
 | [Data][wasi-data]                                                              | Bailey Hayes                           |
 | [proxy-wasm][wasi-proxy-wasm] (will advance as multiple, smaller proposals)    | Piotr Sikora                           |
+| [Networking][wasi-networking]                                                  | Dave Bakker                            |
 
 ### Contributing new proposals
 
@@ -68,3 +69,4 @@ Please see [Contributing to WebAssembly](https://github.com/WebAssembly/WASI/blo
 [wasi-http]: https://github.com/WebAssembly/wasi-http
 [wasi-parallel]: https://github.com/WebAssembly/wasi-parallel
 [wasi-poll]: https://github.com/WebAssembly/wasi-poll
+[wasi-networking]: https://github.com/badeend/WASI-Networking


linclark:
I wanted to talk through the scope of this pre-proposal before merging.

As you point out, networking is a very broad topic, and different protocols make sense for different use cases. Given that, I'm thinking we should split this out into separate proposals, in the same way that we're (retroactively) splitting up `proxy-wasm` into its component parts, like HTTP. With this kind of split between WASI modules, runtimes can make clear which protocols they offer support for and which they don't.

badeend:
> different protocols make sense for different use cases.

Agreed. Even within the same protocol there are different use cases (TCP servers vs TCP clients).

> we're (retroactively) splitting up proxy-wasm into its component parts, like HTTP.

Yes, I've read that in the readme, but I couldn't find any details beyond that statement. Do you have any pointers to where this work is happening?

> Given that, I'm thinking we should split this out into separate proposals. (...) With this kind of split between WASI modules, runtimes can make clear which protocols they offer support for and which they don't.

Seems fine to me. Do you have a specific granularity in mind? (besides a split per protocol)

linclark:
> Yes, I've read that in the readme, but I couldn't find any details beyond that statement. Do you have any pointers to where this work is happening?

On the same proposal page that lists proxy-wasm, you'll see HTTP In Phase 1. This repo has not yet been filled in because it's waiting on the async work that is happening in the component model. If you're interested in the discussions around HTTP and its relationship to async, you can review the WASI meeting minutes in the [meetings repo](https://github.com/WebAssembly/meetings/tree/main/wasi). I also expect that there will be an update on the status of the async work in this week's meeting.

> Seems fine to me. Do you have a specific granularity in mind? (besides a split per protocol)

I don't have any particular split in mind. A good way to start thinking through that might be a discussion in the bi-weekly meeting. If that's something you're interested in, feel free to add an item to the agenda by filing a PR.

badeend:
I've been thinking about what you said and have been reading up on design discussions on the various github repos. I have overhauled the proposal accordingly. TLDR:

- MVP-ified the proposal. Dropped apis that were not strictly neccessary to create a functioning TCP/UDP application.
- Added an api to resolve domain names.
- Moved apis into seperate modules, creating a "layered" structure:
	- mod socket _(Functions applicable to all sockets.)_
		- mod socket_ip _(Functions applicable to Internet sockets. This for example excludes UNIX sockets.)_
			- mod socket_udp
			- mod socket_tcp
	- mod ip_resolve_address
- Clarified what is deliberately not part of the proposal.
- Clarified the security model I have in mind.

I hope the current state of the proposal better reflects the intentions of WASI. If anything, it should be a better starting point for discussion.

I've [requested an agenda item to be added](https://github.com/WebAssembly/meetings/pull/955) for the 2022-02-10 meeting. (I can't make it 2022-01-27)

linclark:
Thanks for taking the time on the updates, and for adding the agenda item. Looking forward to talking about this all more in a couple of weeks!

PR #464, opened 2022-01-07
Fix POSIX equivalent for fd_fdstat_get

turbolent:
It seems like this is a typo


sbc100 (approved):

sbc100:
IIUC this file is auto-generated.  I think you need to update the file which is used to build it: `phases/snapshot/witx/wasi_snapshot_preview1.witx`

sunfishcode:
Thanks!

PR #507, opened 2023-01-18
Remove the handle-index and classic-command proposals.

sunfishcode:
It's now clear that wasi-handle-index is being subsumed by the component model definition of handles, so we no longer need a proposal repository for it.

wasi-classic-command will be replaced by an upcoming command proposal based on the new World concept in the component-model.


sunfishcode:
https://github.com/WebAssembly/WASI/issues/509 tracks the new command World.

PR #517, opened 2023-02-18
Add a "Wit in WASI" document.

sunfishcode:
Add a "Wit in WASI" document explaining how to use Wit in WASI proposals, including some conventions, and the recommended workarounds for component-model features that are not available yet.


sunfishcode:
Since there are no objections so far, I think it makes sense to merge it so that we can write Wit files that reference it. If there are any comments or questions; please file new issues!

PR #522, opened 2023-02-24
Add documentation about starting file descriptors in Preview1.

sunfishcode:
Add documentation mentioning stdin, stdout, and stderr file descriptors, and also preopen file descriptors, and mention that preopen file descriptors can be closed.

Allowing preopen file descriptors to be closed is a change from how Wasmtime historically worked, but I think it's more clear now that it's ok to let preopens be closed. We'll change Wasmtime accordingly.

This addresses WebAssembly/wasi-testsuite#50.

### DIFF
--- a/legacy/application-abi.md
+++ b/legacy/application-abi.md
@@ -52,7 +52,7 @@ with the name `__indirect_function_table`. Function pointers in WASI API calls
 are relative to this table's index space.
 
 When `_start` or `_initialize` is called, environments shall provide file
-descriptors with indices 0, 1, and 2 representing file-like resources for
+descriptors with indices 0, 1, and 2 representing stream resources for
 standard input, standard output, and standard error. Environments may provide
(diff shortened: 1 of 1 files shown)


legacy/application-abi.md
```diff
@@ -51,6 +51,13 @@ Regardless of the kind, all modules accessing WASI APIs also export a table
 with the name `__indirect_function_table`. Function pointers in WASI API calls
 are relative to this table's index space.
 
+When `_start` or `_initialize` is called, environments shall provide file
+descriptors with indices 0, 1, and 2 representing file-like resources for
```
yamt, line 55: what do you mean by "file-like"?
the wording made me think regular files. but i guess it isn't what you meant here.

sunfishcode: That's a good point, "file-like" is ambiguous. I've now changed it to say "stream" instead.

yamt (approved):

PR #523, opened 2023-02-28
Add documentation for return types and pointer types in witx.

sunfishcode:
Document how return types, in particular `expected` return types, work in witx.

Also update the witx-docs.md file to describe the current status of witx and the transition to wit.

### DIFF
--- a/legacy/tools/witx-docs.md
+++ b/legacy/tools/witx-docs.md
@@ -75,8 +75,7 @@ For example, the `poll_oneoff` function has these arguments:
 ```
 
 Pointer values are expected to be aligned, to the alignment of their pointee
-type. If a misaligned pointer is passed to a function, the function shall fail
-with an `errno::inval` error code.
+type. If a misaligned pointer is passed to a function, the function shall trap.
 
 [module linking proposal]: https://github.com/WebAssembly/module-linking/
 [interface types]: https://github.com/WebAssembly/interface-types/blob/main/proposals/interface-types/Explainer.md
--- a/legacy/tools/witx-docs.md
+++ b/legacy/tools/witx-docs.md
@@ -1,30 +1,68 @@
 # Know your `witx`
 
-The `witx` file format is an experimental format which is based on the
-[module linking] text format (`wit`), (which is in turn based on the
-[wat format], which is based on [S-expressions]). It adds some features
-using the same syntax as [interface types], some features with syntax
-similar to [gc types], as well as a few special features of its own.
+`Witx` is an experimental IDL. The text format is based on a text format
+associated with an early version of the [module linking proposal] text
+format, which at the time was called `wit`, though it is a different language
+than what we now call [Wit]. And that `wit` was in turn based on the
+[wat format], which is based on [S-expressions].
 
-`witx` is actively evolving. Expect backwards-incompatible changes,
-particularly in the areas where `witx` differs from `wit`.
+Witx adds some features inspired by [interface types], such as limited
+`string` and `array` arguments, and some features using for working with
+IDL files such as the ability to include one IDL file in another.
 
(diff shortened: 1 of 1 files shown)


sunfishcode:
I've now also added documentation for pointer types, including alignment behavior.

legacy/tools/witx-docs.md
```diff
@@ -1,30 +1,90 @@
...
+For example, the `poll_oneoff` function has these arguments:
+
+```witx
+   (param $in (@witx const_pointer $subscription))
+   (param $out (@witx pointer $event))
+```
+
+Pointer values are expected to be aligned, to the alignment of their pointee
+type. If a misaligned pointer is passed to a function, the function shall fail
+with an `errno::inval` error code.
```
yamt, line 79: i'd suggest s/shall/can/ to align the existing implementations and wasm `load/store` semantics.

legacy/tools/witx-docs.md
```diff
@@ -1,30 +1,90 @@
...
+The WASI Subgroup is [migrating] away from `witx` and toward [Wit], because
+Wit provides much better support for non-C-like languages, better support for
+API virtualization, it has a path to integrating async behavior into WASI
+APIs in a comprehensive way, and it supports much more expression APIs, such
+as the ability to have `string`s and other types as return types in addition
+to just argument types. At this point, the tooling for Wit is also a lot more
+sophisticated and the [Wit language] and [Canonical ABI] have much more
+documentation.
+
+This document focused on the witx format.
```
abrown, line 26:
```suggestion
This document is focused on the witx format.
```

abrown (approved):

legacy/tools/witx-docs.md
yamt: well, actually i prefer to just accept misaligned addresses as wasm `load/store` does.

even if we choose to reject them, i think it should be with a trap (as in eg. [wasm threads](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#alignment)), not `inval`.

sunfishcode: That sounds good to me. I've now changed it to say trapping.

legacy/tools/witx-docs.md
```diff
@@ -1,30 +1,89 @@
...
-WebAssembly standards and proposals wherever practical, and to be an input 
-into the design process of interface types.
-
-[module linking]: https://github.com/WebAssembly/module-linking/blob/main/design/proposals/module-linking/Explainer.md
+to existing APIs and proposing new APIs.
+
+The WASI Subgroup is [migrating] away from `witx` and toward [Wit], because
+Wit provides much better support for non-C-like languages, better support for
+API virtualization, it has a path to integrating async behavior into WASI
+APIs in a comprehensive way, and it supports much more expression APIs, such
```
yamt, line 20: expressive?

legacy/tools/witx-docs.md
yamt: will preview2 have the same trapping semantics?

yamt (commented):
lgtm

legacy/tools/witx-docs.md
sunfishcode: Yes; as currently proposed, the [CanonicalABI has](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#loading) traps for both alignment and bounds.

legacy/tools/witx-docs.md
sunfishcode: yep; fixed

sunfishcode:
Sounds good; let's merge this, and if anyone has further questions, we can iterate in new issues and PRs.

PR #530, opened 2023-04-02
Fix broken link wasi-digital-io

emielvanseveren:
The link to wasi-digital-io now goes to correct repository.


sunfishcode:
Thanks!

PR #531, opened 2023-04-07
Allow file timestamps to be zero

yamt:
cf.
https://github.com/bytecodealliance/wasmtime/pull/4531
https://github.com/WebAssembly/wasi-testsuite/pull/59


sunfishcode:
This makes sense, and there are no objections here, so let's do this.

PR #549, opened 2023-07-24
Rebase the Phase Process description on the CG's current process

sunfishcode:
The [CG Phase Process document] has recently split out the entry requirements for each stage from the activities that happen within each stage, fixing an ambiguity about what happens before a stage and what happens within a stage. It also contains a number of generally useful updates.

This PR updates the WASI Phase Process using wording derived from the CG Phase Process, adapting it to meet WASI's needs. The resulting process is roughly the same as the existing process, however I've made it more specific in a few areas:

 - Phase 2 requires a wit description.
 - Phase 3 requires there be a plan for how the phase 4 accpetance criteria will be met.

[CG Phase Process document]: https://github.com/WebAssembly/meetings/blob/main/process/phases.md

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -3,58 +3,107 @@
 Interested in participating? Please follow
 [the same contributing guidelines as the design repository][].
 
-  [the same contributing guidelines as the design repository]: https://github.com/WebAssembly/design/blob/master/Contributing.md
+[the same contributing guidelines as the design repository]: https://github.com/WebAssembly/design/blob/master/Contributing.md
 
 Also, please be sure to read [the README.md](README.md) for this repository.
 
-## Championing a Proposal
+To contribute to an [existing proposal](https://github.com/WebAssembly/WASI/blob/main/Proposals.md),
+refer to the linked proposal repository.
 
-If you want to champion a new proposal, here's what you need to do in each phase:
+The start a new proposal, the first step is to file an issue in the
+[WASI repository](https://github.com/WebAssembly/WASI/issues) presenting
+the idea. A good API proposal should discuss the scope of the API,
+the use cases, and the places it would be expected to be implemented.
+Then proceed with the rest of the steps in phase 0 described below.
 
-### Phase 0: Gauge interest
+If you have any questions about any step of the process, please reach out
+to one of the WASI Subgroup (SG) chairs.
 
-You have an idea for an API. To see whether others are interested in pursuing the idea, you should work up a rough description of the API and post it somewhere that is publicly visible. This could be in the WASI issue queue, or in a gist or as its own repo. You can use the [proposal template] if you like, but it's not required in this phase.
+## The Phase Process
 
-Once you've done this, you can optionally have the subgroup discuss the idea by adding a discussion item to the [WASI meeting agenda].
+The following process is modeled after [WebAssembly CG's Phase Process],
+though it differs in several areas, to reflect the unique needs of APIs.
 
-Once you feel ready, you can add a vote to the [WASI meeting agenda] to move to the next stage.
+Something is out-of-scope if it doesn't fit the [WASI Subgroup's charter](https://github.com/WebAssembly/WASI/blob/main/Charter.md) and there's agreement that the charter should not be amended to cover the proposal.
 
-### Phase 1: Write spec text
+In general, the process moves forward through a series of numbered phases.
+However, if issues are uncovered or consensus devolves,
+proposals should back up to the appropriate prior step.
 
-At this point, the WASI SG chair will create a new repo for the proposal in the WebAssembly GitHub org. This will follow the conventions of the [proposal template] and [Wit in WASI](docs/WitInWasi.md). If you have any questions about how to fill in the spec template, you can reach out to the WASI SG chair.
+No vote is required for a proposal to enter phase 0. To advance from one phase
+to another, a vote proposing the advancement is added to a
+[WASI Subgroup meeting](https://github.com/WebAssembly/meetings/tree/main/wasi) agenda
+through a pull request, and the SG votes on whether to approve it, evaluating
+whether the new phase's entry requirements have been met.
 
-As part of moving to the next phase, the champions need to define the acceptance criteria for Phase 4. This is because WASI includes APIs that cover a diversity of different domains and use cases, so the acceptance criteria can be very different between different proposals.
+### 0. Pre-Proposal [Individual Contributor]
 
-Some examples of potential criteria:
+Entry requirements:
 
-- multiple independent production implementations
-- implementations on multiple host platforms
-- polyfill implementations
-- bindings in toolchains and libraries
+  * A WASI Subgroup (SG) member has an idea. Notably, no SG vote is required to begin phase 0.
 
-Note: portability requirements may vary between proposals, as not all features will necessarily make sense in all host environments.
+During this phase:
 
-With all this in place, you can add a vote to [WASI meeting agenda] to move to the next stage.
+  1. An issue is filed on the [WASI repository](https://github.com/WebAssembly/WASI/issues) to present the idea.
+  1. Discussion on the API occurs on the issue.
+  1. A champion or champions emerge. They may add the proposal to the [proposal list](https://github.com/WebAssembly/WASI/blob/main/Proposals.md) at phase 0.
+  1. The champion(s) put together a description of the API in their own GitHub repository or on the issue. You can use the [proposal template] if you like, but it's not required in this phase.
 
-### Phase 2: Work with implementers to prototype and refine the design
+### 1. Feature Proposal [WASI Subgroup]
 
-At this point, you should be prototyping the API to make sure it works in practice, and you should develop a test suite which can be used by other implementations to validate their spec compliance.
+Entry requirements:
 
-Once the implementation has stabilized, it's again time to add a vote to [WASI meeting agenda] to move to the next stage.
+  * There is general interest within the SG in this API.
+  * The CG believes the API is in-scope and will plausibly be workable.
 
-### Phase 3: Validate the design through multiple implementations
+During this phase:
 
-At this point, you'll need to get more implementers involved. How many implementations you need depends on the Phase 4 acceptance criteria that you set in Phase 2.
+  1. If the proposal is not already listed, it should be added to the [proposal list](https://github.com/WebAssembly/WASI/blob/main/Proposals.md) at this time.
+  1. A new repository, forking the [proposal template] repo, is created by one of the SG chairs, or transferred to the WebAssembly organization by the champion.
+  1. The champion will attempt to reach broad consensus in the Subgroup.
+  1. Pull requests and issues are used to iterate on the design of the API. Specifically, an overview document must be produced that specifies the API with reasonably precise and complete language before attempting to move to phase 2 (meaning it is sufficiently precise to be implemented following this description, without obvious holes or ambiguities).
+  1. If relevant to demonstrate the viability of a API, prototype implementations of the API are implemented by interested embedders (possibly on a branch).
 
-You may need to make changes in response to implementer feedback, but we expect the API to be pretty stable by this point. If implementors uncover especially challenging design issues, the proposal may be sent back to Phase 2 for more development.
+Additionally during this phase:
 
-Once the implementations are in place, you can add the final WASI SG vote to [WASI meeting agenda]. After this, the proposal advances to a vote in the broader WebAssembly CG.
+ * The champions define the acceptance criteria for Phase 4.
 
-### Phases 4 & 5: Push it over the finish line
+   This is intended to translate the spirit of the CG Phase Process' "Two or more Web VMs" requirement to meet WASI's needs. The criteria should establish at least:
+    - Portability: WASI APIs should be portable, however that can mean different things to different use cases, and no one definition covers everything. Consequently, each proposal should define criteria establishing its specific portability requirements.
+    - Practicality: It's important that WASI APIs be implementable and usable in real-world use cases, so each proposal should define criteria establishing a sufficient level of confidence.
+    - Testing: APIs will have different needs in terms of environments needed to test them, so each proposal should define criteria establishing what form the testing will take.
 
-The specific process in Phases 4 and 5 will be determined when we have a proposal ready for them.
+### 2. Feature Description Available [WASI Subgroup]
 
-Note: While we mostly follow the [WebAssembly CG's Phase Process], the requirements around Web VM implementation, formal notation and the reference interpreter don't apply in the context of WASI.
+Entry requirements:
+
+   * The phase 4 acceptance criteria are documented in the proposal.
+   * Precise and complete overview document is available in a proposal repo around which a reasonably high level of consensus exists.
+   * A [wit](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) description of the API exists.
+   * All dependencies of the wit description must have reached phase 2.
+
+During this phase:
+
+   * One or more implementations proceed on prototyping the API.
+   * A plan is developed for how the phase 4 acceptance criteria will be met.
+
+## 3. Implementation Phase [WASI Subgroup]
+
+Entry requirements:
+
+   * The phase 4 acceptance criteria must be either met or there must be a plan for how they're expected to be met.
+   * All dependencies of the wit descriptions must have reached phase 3.
+
+During this phase, the following proceeds in parallel:
+
+   * Implementations are built
+   * Toolchains, libraries, and other tools using the API are built
+   * Remaining open questions are resolved.
+   * The plan for satisfying the phase 4 acceptance criteria is followed, though the plan may change over time.
+
+### Phases 4 & 5: To be determined
+
+Phases 4 and 5 are where a feature is finished and standardized. As WASI matures, the WASI Subgroup will coordinate with its parent WebAssembly Community Group and the WebAssembly Working Group to define a process for standardization.
 
 [proposal template]: https://github.com/WebAssembly/wasi-proposal-template
 [WASI meeting agenda]: https://github.com/WebAssembly/meetings/tree/main/wasi
--- a/Contributing.md
+++ b/Contributing.md
@@ -69,6 +69,7 @@ Additionally during this phase:
  * The champions define the *portability criteria* for Phase 4.
 
    This is intended to translate the spirit of the CG Phase Process' "Two or more Web VMs" requirement to meet WASI's needs. The criteria should establish at least:
+    - Two or more implementations: Each proposal should say what kinds of implementations.
     - Portability: WASI APIs should be portable, however that can mean different things to different use cases, and no one definition covers everything. Consequently, each proposal should define criteria establishing its specific portability requirements.
     - Practicality: It's important that WASI APIs be implementable and usable in real-world use cases, so each proposal should define criteria establishing a sufficient level of confidence.
     - Testing: APIs will have different needs in terms of environments needed to test them, so each proposal should define criteria establishing what form the testing will take.
--- a/Contributing.md
+++ b/Contributing.md
@@ -66,7 +66,7 @@ During this phase:
 
 Additionally during this phase:
 
- * The champions define the acceptance criteria for Phase 4.
+ * The champions define the *portability criteria* for Phase 4.
 
    This is intended to translate the spirit of the CG Phase Process' "Two or more Web VMs" requirement to meet WASI's needs. The criteria should establish at least:
     - Portability: WASI APIs should be portable, however that can mean different things to different use cases, and no one definition covers everything. Consequently, each proposal should define criteria establishing its specific portability requirements.
@@ -77,7 +77,7 @@ Additionally during this phase:
 
 Entry requirements:
 
-   * The phase 4 acceptance criteria are documented in the proposal.
+   * The portability criteria are documented in the proposal.
    * Precise and complete overview document is available in a proposal repo around which a reasonably high level of consensus exists.
    * A [wit](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) description of the API exists.
    * All dependencies of the wit description must have reached phase 2.
@@ -85,13 +85,13 @@ Entry requirements:
 During this phase:
 
    * One or more implementations proceed on prototyping the API.
-   * A plan is developed for how the phase 4 acceptance criteria will be met.
+   * A plan is developed for how the portability criteria will be met.
 
 ## 3. Implementation Phase [WASI Subgroup]
 
 Entry requirements:
 
-   * The phase 4 acceptance criteria must be either met or there must be a plan for how they're expected to be met.
+   * The portability criteria must be either met or there must be a plan for how they're expected to be met.
    * All dependencies of the wit descriptions must have reached phase 3.
 
 During this phase, the following proceeds in parallel:
@@ -99,7 +99,7 @@ During this phase, the following proceeds in parallel:
    * Implementations are built
    * Toolchains, libraries, and other tools using the API are built
    * Remaining open questions are resolved.
-   * The plan for satisfying the phase 4 acceptance criteria is followed, though the plan may change over time.
+   * The plan for satisfying the portability criteria is followed, though the plan may change over time.
 
 ### Phases 4 & 5: To be determined
 
--- a/Contributing.md
+++ b/Contributing.md
@@ -54,7 +54,7 @@ During this phase:
 Entry requirements:
 
   * There is general interest within the SG in this API.
-  * The CG believes the API is in-scope and will plausibly be workable.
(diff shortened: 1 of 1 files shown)


abrown:
From the PR, for phase 2:

> A [wit](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) description of the API exists.

Several recent conversations have made me pause and consider what this change might do to the ecosystem. Not all of the implications of this change are clear to everyone and I think we should make them clear here to avoid discontent in the future. To my eyes, the WebAssembly ecosystem is already quite fractured (have you tried building a module that runs on a standalone engine AND the web?) and one of the bright spots was the agreement from several different corners of the web on the WASI standard. I've heard the concern that binding WASI to WIT &mdash; and implicitly to the component model &mdash; might cause fractures (e.g., other standards or non-standard APIs).

Let me propose some questions so that someone can clarify where all this is headed. I'll do it from various perspectives:

1. _For the engine implementers_: do I have to implement the component model to continue supporting WASI?

2. _For WASI API designers_: what will I do if I can't express my API in WIT?

3. _For WASI API users_: will code that I compiled to use WITX-defined WASI APIs (not just `preview1`) work in component model engines? What about the reverse &mdash; code using WIT-defined APIs in non-component model engines?

4. _For performance optimization_: will every call across the WIT boundary incur data copy overhead?

I think I know the answer to some of these questions and some of them have been discussed in the past (cc: @sunfishcode, @pchickey) but I think making all of this explicit will be helpful.

xwang98:
We wish the component model is not mandatory for WASI. Component model introduce complexity and additional resource requirement, and it is not always wanted. For exmple the footprint is a hard requirement for embedded and IoT usages, we wish we can still use wasi for these domains.

yamt:
my understanding is that having a WIT-defined interface doesn't imply to require component-model support.
you can use the core-wasm level abi directly if you want.

my impression is that WIT-defined interfaces are often less efficient if you compare them with an abi based on bare linear memory pointers though.

woodsmc:
I agree, with @abrown here answers to these questions would be helpful for many in the ecosystem. To @abrown's observation, the answers to these questions are often known inside the team working on the Component Model. It's just a change in perspective; approaching the Component Model from the outside in.

I know many folks may be worried about stating known limitations. But I think it's an opportunity for community engagement. For instance, I'm aware that there is scope of performance improvements in how the marshalling between components work.  For those enthusiastic about the component model it points out an area where contributions and community focus may be welcome.

On the general topic of performance and overhead; To @xwang98's point, we have similar concerns, but feel that getting some hard data on the performance and impact would be great.  It would allow everyone to assess it's suitability for their particular use cases. We'd love to see some performance metrics / data. But being realistic - I also know this isn't going to be possible until after preview 2 is released. Again, stating this as an area of contribution or focus, following the release of Preview 2 would be great.

Regarding the engine impact, would it be possible to get some some engineering guidance from those that have implemented the component model in an engine already - I'm guessing this may be the Wasmtime team? This would help address the concerns of other runtimes and provide guidance on the suitability of the component model for various domains. Maybe a future blog post or interview? - just a thought.

tschneidereit:
> I think I know the answer to some of these questions and some of them have been discussed in the past (cc: @sunfishcode, @pchickey) but I think making all of this explicit will be helpful.

Thank you for making raising these questions this explicitly. I agree that it makes sense to explicitly work through them for other interested parties, so I'll walk through them in way more detail than you personally would need.

(Note: this got very long, and I apologize. **I'd highly recommend reading the first section**, and then those Q&A entries you're interested in.)

One thing I want to emphasize is that

### **nothing has changed about any of this in a long time!**

The particular phase 2 requirement you mention ("A [wit](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md?rgh-link-date=2023-08-03T18%3A42%3A04Z) description of the API exists") has been in place since 2021, and explicitly and fairly prominently mentioned in the main README [since early February 2022](https://github.com/WebAssembly/WASI/pull/451/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R13).

The more fundamental approach of basing WASI on top of another standard to define the ABI is even older: [the very first WASI overview document](https://github.com/WebAssembly/WASI/blob/b81f586ecdca184764e91587a18d59b1d85242c9/docs/WASI-overview.md#portable-system-interface-for-webassembly) from when WASI was announced in April 2019 mentions WASI gaining support for "Host Bindings". Since then, the Host Bindings proposal merged with the Module Linking proposal into what's now the Component Model.

That very first WASI overview also already includes the reason for moving towards defining WASI in terms of Host Bindings: the ABI approach used by WASI Preview 1 works very well for languages like C, C++, and Rust, which use linear memory. It doesn't work well at all for languages like Kotlin and Dart, which use Wasm GC. That's because it fundamentally assumes that there is a linear memory heap to read values from and write them into.

With the Component Model's approach of defining the ABI in terms of canonical lowering and lifting operations for each data type, we can directly and efficiently support languages using GC by defining lifting and lowering operations that operate on GC objects. 

Based on this, another thing I want to highlight is that

### **not using an approach like the Component Model's WIT-based APIs means no real support for Kotlin, Dart, and other languages using Wasm GC**

In summary, this PR clarifies some aspects of the process, but doesn't in any way represent a change in direction.

@xwang98 the above means that the direction WASI is on hasn't changed, and that being based on what is now the Component Model has been part of the design since the very beginning. This approach has since been confirmed a number of times, and changing it now would not only mean discarding many person-years of work, but also require coming up with a different approach to at least some of the goals. E.g. I'm sure you agree that not supporting languages using Wasm GC isn't really an option.

With that all out of the way, I'll try to answer @abrown's questions below, as well as some of the concerns @woodsmc raised.

-----

### Q&A

> 1. _For the engine implementers_: do I have to implement the component model to continue supporting WASI?

As @yamt says: no, that's not required to support content that'd otherwise use a WASI Preview1-style ABI using WITX. WebAssembly Components define a new binary format, but to support content that works in roughly the same way as Wasm core modules targeting Preview1, all that's needed is the ability to "unwrap" the core modules contained in these components, and then communicating with those via  the [canonical ABI](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md). This ABI is roughly equivalent to the ABI witx defines and which is used in WASI Preview1. This approach is e.g. taken by the [JCO toolchain](https://github.com/bytecodealliance/jco) to support running components in JS engines such as browsers or Node.js which (for now) lack native support.

Another strong proof that this works is that we have multiple toolchains able to produce Components, despite none of them actually emitting the new binary format. Instead, they all emit core Wasm modules with the Component Model's ABI, which are then turned into Component binaries using external tooling.

> 2. _For WASI API designers_: what will I do if I can't express my API in WIT?

WASI has as part of its fundamental design goals extremely high security standards, the ability to treat all languages as first-class citizens—instead of just ones that behave like C—and the ability to make all APIs fully virtualizable. (I.e., to enable all APIs to be implemented in WebAssembly and with the same privileges of all other content.)

Based on these goals, WASI (by virtue of being based on the Component Model) introduces two major constraints that can impose limitations on some API designs:
1. full encapsulation of a Component's internals, with the WIT-defined API it exposes being the only way to interact with it
2. a limitation to host calls that can be implemented in content instead of just the host

I don't think that any of this means that there is any kind of functionality that WIT fundamentally can't expose an API for. It's true however that some API designs won't work, and will need other approaches.

I think the most important reason for that is the need for all APIs to be language-agnostic though. E.g. shared-everything multi-threading doesn't really mean the same thing for languages using linear memory and those using Wasm GC. Features of that nature are I think best handled as Component-internal, much like Wasm GC itself, exception handling, or stack switching.

> 3. _For WASI API users_: will code that I compiled to use WITX-defined WASI APIs (not just `preview1`) work in component model engines? What about the reverse — code using WIT-defined APIs in non-component model engines?

Content will run in those runtimes that support the ABI and binary format it uses, and that implement the APIs it requires. There's absolutely nothing stopping runtimes from supporting all kinds of different ABIs and binary formats, so the answer to the first question is definitely "yes".

Additionally, it's possible to fully support WASI Preview1-targeting content *inside* the component model. The Wasmtime project has been working on an [adapter](https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi-preview1-component-adapter) for just that, which should enable all Preview2-supporting runtimes to support Preview1 without maintaining multiple WASI implementations in parallel.

Since WASI Preview2 introduces a whole host of additional functionality, such as the [wasi-http API](https://github.com/WebAssembly/wasi-http/), it's not really possible to support Preview2 in runtimes that don't support these interfaces. But as mentioned above, runtimes can choose to only implement support for the canonical ABI and running single modules, instead of supporting linked Components as well.

> 4. _For performance optimization_: will every call across the WIT boundary incur data copy overhead?

Not in any way that's not there for WITX as well, and in fact inherent to WebAssembly in general. The fact that content runs inside a tight sandbox means that one can't just expose arbitrary regions of memory and operate on it without copying.

The Component Model does however introduce a concept that significantly reduces this overhead: [resources and resource handles](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-resource). These can represent a large collection of values without having to copy them, and enables operating on them via associated functions/methods.

-----

@woodsmc, I hope I was able to address some of your concerns with the above, but I'm very happy to discuss things in more detail (though maybe not as part of a PR that's not actually about any of this 😉)

woodsmc:
Thanks @tschneidereit . 

>Thank you for making raising these questions this explicitly. I agree that it makes sense to explicitly work through them for other interested parties, so I'll walk through them in way more detail than you personally would need.

Awesome; I know this is a perspective shift - rather than addressing our own internal community, we are answering the anticipated questions of others. I love this. Mainly because I get asked these types of questions from my organization frequently. Having them explicitly addressed lowers the barrier for entry and aids with technology adoption. 

It would be amazing to have these points addressed explicitly in some "published" form, an addition to this document, or another.  So thanks again!

RE: Performance / other issues - delighted to take it out of the PR. I'll reach out. Thank you.

Contributing.md
```diff
@@ -3,58 +3,107 @@
...
+    - Practicality: It's important that WASI APIs be implementable and usable in real-world use cases, so each proposal should define criteria establishing a sufficient level of confidence.
+    - Testing: APIs will have different needs in terms of environments needed to test them, so each proposal should define criteria establishing what form the testing will take.
 
-The specific process in Phases 4 and 5 will be determined when we have a proposal ready for them.
+### 2. Feature Description Available [WASI Subgroup]
 
-Note: While we mostly follow the [WebAssembly CG's Phase Process], the requirements around Web VM implementation, formal notation and the reference interpreter don't apply in the context of WASI.
+Entry requirements:
+
+   * The phase 4 acceptance criteria are documented in the proposal.
```
penzn, line 80: Is it meant to be phase 3? There is a reference to the acceptance criteria there, while phase 4 seems to be "TBD"

Contributing.md
```diff
@@ -3,58 +3,107 @@
...
+During this phase:
+
+   * One or more implementations proceed on prototyping the API.
+   * A plan is developed for how the phase 4 acceptance criteria will be met.
+
+## 3. Implementation Phase [WASI Subgroup]
+
+Entry requirements:
+
+   * The phase 4 acceptance criteria must be either met or there must be a plan for how they're expected to be met.
```
penzn, line 94: Should this be phase 3?

penzn (commented):
There is some inconsistency of where acceptance criteria is checked, maybe needs some clarification.

Contributing.md
sunfishcode: Thanks for bringing this up.

The "phase 4 acceptance criteria" are [defined here](https://github.com/WebAssembly/WASI/blob/main/Contributing.md#phase-1-write-spec-text), and are attended to be a substitute for the "two Web VMs" requirement in the corresponding phase 4 in the  CG process.

Phase 4 in the CG process means "the feature is done and handed off to the WG for proper standardization". For WASI, we aren't ready for anything analogous to that yet; we're still working on what we're calling "Preview 2".

So, this PR proposes defining "Preview 2" as requiring proposals be at only phase 3. But, we do want Preview 2 proposals to have something of the level of confidence that "Two Web VMs" have, so we added this language about the "phase 4 acceptance criteria" as a Preview 2 requirement.

Does that clarify it?

Contributing.md
sunfishcode: It's written as intended; see my comment above.

Contributing.md
penzn: Maybe it's just me, but currently "phase 4 acceptance criteria" sounds like "acceptance criteria to a stage we haven't yet defined", also logically requiring phase 4 before phase 3 doesn't make very strong sense. Maybe have the criteria point to phase 3 officially or give them a phase-neutral name (how about "Preview 2 acceptance criteria", since that is the end goal)? Or define a phase 4?

sunfishcode: If we renamed the "phase 4 acceptance criteria" to the "portability criteria", would that make it clear?

penzn: It should be more clear that way, I think

sunfishcode: I've now performed the rename, here and in #550.

Wrt your edit, I've also now re-introduce "two or more implementations" language. That was previously only an example, but it does seem worth having, even if we have to leave it up to proposals to define what kinds of implementations are needed.

yamt:
> This ABI is roughly equivalent to the ABI witx defines and which is used in WASI Preview1. 

actually, it's sometimes considerably more expensive as it involves malloc.

yamt:
> RE: Performance / other issues - delighted to take it out of the PR. I'll reach out. Thank you.

if you moved the discussion to elsewhere, give me a pointer to the new place. thank you.

sunfishcode:
@yamt The places where the ABI does a malloc fall into two categories: there are some malloc calls that we haven't yet optimized yet but will, and there are some malloc calls in areas that have no witx equivalent.

yamt:
> @yamt The places where the ABI does a malloc fall into two categories: there are some malloc calls that we haven't yet optimized yet but will, and there are some malloc calls in areas that have no witx equivalent.

which category does eg. fd_read (besides malloc, it lacks of iov) fall into?

yamt:
> @yamt The places where the ABI does a malloc fall into two categories: there are some malloc calls that we haven't yet optimized yet but will, and there are some malloc calls in areas that have no witx equivalent.

when you say "haven't yet optimized yet but will", do you mean adapter functions?

sunfishcode:
> which category does eg. fd_read (besides malloc, it lacks of iov) fall into?

We haven't optimized it yet.

> when you say "haven't yet optimized yet but will", do you mean adapter functions?

The malloc call can be optimized away by changing how the bindings are generated. iov functionality would require adding the feature to the canonical ABI spec, but it's doable.

yamt:
> > which category does eg. fd_read (besides malloc, it lacks of iov) fall into?
> 
> We haven't optimized it yet.
> 
> > when you say "haven't yet optimized yet but will", do you mean adapter functions?
> 
> The malloc call can be optimized away by changing how the bindings are generated. iov functionality would require adding the feature to the canonical ABI spec, but it's doable.

ok.

do you have any idea when/if such optimizations can be made?

at core wasm level, how do such optimized versions look like?
will they have different import names from the current version?
i guess it's difficult to distinguish optimized and current versions by core func type alone.

penzn:
Recap of some of the discussions that I think would be useful to share:

- Restrictions this approach represents preclude wasi-threads from being officially included in any of WASI preview versions
- WebAssembly/component-model#239, when we had this conversation it sounded like static core linking would happen at runtime, trying to make sure I got that right

woodsmc:
In the interests of visibility I wanted to share and get some feedback on two key items, which should be considered in relation to this PR, that is *performance equivalence* and *the parallel life of WIT and WITX*. Both of these help to address the concerns expressed in this thread and related discussions.

**Performance Equivalence**  
Could we change the PR, to include a stipulation that the WASI interfaces which currently exist shouldn't see a performance degradation?...  This was discussed with @lukewagner @ricochet. The performance criteria would allow for variation and innovation of the implementation, and indeed the standard going forward while ensuring no negative impact for existing solutions.

**Justification / Mitigation of Concern**  
The WIT format is essentially a moving target at the moment. It is continuing to evolve. New concepts are already planned for introduction. An `async` concept has been proposed. This is an example of a primitive which is exclusive to a set of more modern languages, like Rust. `async` doesn't exist in C .

A concern would be these newer language specific primitives being used to define interfaces upon which existing C code is heavily dependent. A socket interface is a great example of this. If a socket interface were to be implemented using `async` then it would work great for Rust, but would suck for C, and other non-`async`y languages. Today, in the C world, we'd need to write a bunch of code to turn that asynchronous function into a synchronous one. This could foreseeably result in C applications under-performing on newer WASI system calls - disproportionately impacting them.  

At the same time it is important not to rule out innovation, and technology advancement, both in the runtime implementation and in the realization of the WASI standards themselves.   So rather than blanket rejecting of new concepts, like `async`, the introduction of performance check means that we can allow the implementation and standard to evolve over time, while ensuring that we don't disproportionally affect individual languages, particularly C, which is critical for Industrial and IoT based applications.

The implication of course, may be, with regard to the malloc discussion between @sunfishcode and @penzn that the adoption of a WIT implementation would be dependent on showing no regression on performance between a witx implementation and a wit implementation. Some data here to validate this may be required?

The resulting performance checks, may result in performance observations in a number of the key languages WASM targets. C, Rust, Go ? - Others ?

I know this puts an additional onus on those proposing standard changes, and that this may slow standard evolution, but this, perhaps is justified, since as the WASI ecosystem matures the cost of interface change increases, and rapid, performance impacting changes will drive away adoption in general.

**The Parallel life of WIT and WITX**  
During the discussion between @lukewagner and @xwang98 a recommendation was made that while WIT may be used for standardization and discussion, that it should be possible, using a tool to derive witx, at least until a point in the future where we have support for WIT and the component model in a wider set of runtime implementations. 

**Justification / Mitigation of Concern**  
This provides the necessary time to allow some of the embedded customers of WASM to plan and adopt a migration strategy. While also allowing them to continue to propose, adopt, and prepare for new interfaces and proposals.

This may extend the life of WITX; but it provides an important runway for WAMR in particular. There are WAMR users with 100,000s of individual devices deployed in the field. The rate of runtime change is considerably slower in the IoT and embedded world, than in the cloud / data center environment. Aside from the practical implications of deploying the an updated runtime, there is a need to allow time for the runtime's own evolution to support the component model, and to ensure it can continue to execute the same functional payload with the existing hardware specifications (see, performance above).

A  concern would also be that a move to a WIT only world at this point in the WASI journey, may result in standards only being proposed and considered by the subset of the community actively engaged in WIT compliant runtimes. As those not working on runtimes which can adopt the standards would struggle with the context necessary for active participation in the conversation, and wouldn't be able to propose standards which they themselves could implement.... at least until we've wider support... 

Taken together, I think these two suggestions help to address a number of concerns which impact the embedded and IoT world. Thoughts?

penzn:
My final 2¢, I don't meant to be gloomy or delay adoption of Component Model in WASI, I just think this should be shared for awareness.

I think this change, with mandating component model, as opposed to being 'just an external API', presents a qualitative change, however minor, where WASI would need some core features. There are two reasons for it in my view:

- Component Model itself is a core proposal, at least far as APIs are concerned
- Certain features must be implemented in core wasm instead of APIs, as we have seen with [threads](https://github.com/WebAssembly/wasi-threads/issues/48#issuecomment-1696407630)

This is concerning, because so far there is not much traction in supporting component model or other core features mandated by this change in browsers (treads draft hasn't been presented yet, for example). This would create divergence between WASI and 'stock' web environment, while _more_ convergence would be preferrable in my opinion, as there are currently [some challenges](https://code.visualstudio.com/blogs/2023/06/05/vscode-wasm-wasi#_vs-codes-wasi-implementation) with running the same code both ways.

Maybe a compromise would be to allow WIT and WITX coexist for the time being, at least this way exploration of component model can continue while maintaining backwards-compatibility with the existing WASI approach.

lukewagner:
(sorry for the slow reply due to holiday + travel) Thanks for the thorough writeup @woodsmc. For my part, I agree on both your broader points and suggestions. As for how I think we could go about concretely integrating these into the docs and process:

For the ‘Performance Equivalence’ point, while I agree that our goal is to ensure that Preview 2 doesn't regress performance (after all, performance is one of the main motivating factors for using wasm in the first place), I think it’s important that we don’t choose a fragile evaluation criteria or one that prevents iteration and real-world feedback. Having worked on benchmarks for some years before, it’s surprisingly easy to write well-meaning benchmarks that completely misrepresent real-world performance and lead developers in the wrong direction. Additionally, just by nature of older code paths having received more optimization, I think it’s important to distinguish between any temporary regressions that may occur due to newness and lack of optimization vs. essential performance differences that indicate roadblocks to better performance in the future. Lastly I think, at this point in time, it's really important to ship our next iteration of WASI this year and fix any lingering performance issues in the next iteration next year. Based on all that, my suggestion is that we add a “Performance goals” subsection to preview2/README.md (added in #550) that says something to the effect of:
  > It is a goal and expectation that Preview 2 will not regress performance of real-world workloads compared to Preview 1.  In the Preview 2 timeframe, WASI interfaces will aim to achieve the best performance they can given the current feature set of Wit and the Component Model.  However, in the interest of shipping Preview 2 promptly to gather real-world feedback and inform the next iteration (Preview 3), modest regressions may need to wait until the next iteration to be fully addressed.

For the ‘Parallel Life of WIT and WITX’ point, agreed and perhaps we could add a “WITX” section to the current [WitInWasi.md](https://github.com/sunfishcode/WASI/blob/sunfishcode/proposal/docs/WitInWasi.md) that describes how .witx files can be derived from .wit files according to the Canonical ABI and how wasm engines can implement single-module components using just these derived .witx files and their existing WITX machinery.

Does that sound reasonable?

lukewagner:
@penzn Agreed that we should generate WITX from WIT (via the Canonical ABI) to help developers transition.  However, to the wasi-threads point, I think threads are wholly independent: the problem with wasi-threads is the O(MxN) function table space usage it implies (assuming M threads and N functions) and the consequent problems for dlopen()-style dynamic linking, not the component model.  If it weren't for these problems, it would be easy enough to add a `thread.spawn` [canon built-in](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#canonical-built-ins) to the component model (there are a few already).  Thus, we shouldn't consider the component model as part of the threading story -- that is orthogonal and driven by fundamental core wasm representation issues.

penzn:
> I think threads are wholly independent: the problem with wasi-threads is the O(MxN) function table space usage it implies (assuming M threads and N functions) and the consequent problems for dlopen()-style dynamic linking, not the component model.

Do you have a link to a discussion/notes where this was stated previously? I can see a couple prior discussions where it wasi either implied or directly said that threads and similar APIs are ultimately incompatible with Component Model. For example in https://github.com/WebAssembly/wasi-threads/issues/48#issuecomment-1644487201: `These limitations also turn out to make it incompatible with the component model and WASI Preview 2`, and further down. Further, function references/pointers can't be represented in component model syntax (WebAssembly/component-model#117) and those would be required for a threads-like API.

Even if the issue of threads was completely decoupled from component model, then it would be an additional instance of needing core a wasm feature in WASI, [instance-per-thread multithreading](https://emscripten.org/docs/porting/pthreads.html) is used on the Web today.

yamt:
> However, to the wasi-threads point, I think threads are wholly independent: the problem with wasi-threads is the O(MxN) function table space usage it implies (assuming M threads and N functions) and the consequent problems for dlopen()-style dynamic linking, not the component model.

component model is often used to link core instances and it in some cases needs to involve function table, thus has problems wrt wasi-threads similar to dynamic linking, doesn't it?

yamt:
> > > which category does eg. fd_read (besides malloc, it lacks of iov) fall into?
> > 
> > 
> > We haven't optimized it yet.
> > > when you say "haven't yet optimized yet but will", do you mean adapter functions?
> > 
> > 
> > The malloc call can be optimized away by changing how the bindings are generated. iov functionality would require adding the feature to the canonical ABI spec, but it's doable.
> 
> ok.
> 
> do you have any idea when/if such optimizations can be made?
> 
> at core wasm level, how do such optimized versions look like? will they have different import names from the current version? i guess it's difficult to distinguish optimized and current versions by core func type alone.

my understanding is that such an optimization will be an ABI breaking change at core wasm level.
so, if we allow to implement apis at core wasm level (w/o component model), we need some kind of versioning visible at core wasm level, like somehow changing import/export names.
right?

sunfishcode:
I've now added commits to #550 to add the content of the discussion about [performance](https://github.com/WebAssembly/WASI/pull/550/commits/778c9310012b5525f00c7d7cd75d83f52031ccac) and [witx and wit](https://github.com/WebAssembly/WASI/pull/550/commits/25e8370173de64a893b1ff74d8a41ccc39660a17).

pchickey:
This PR was approved by unanimous consent in the 9/07/23 WASI subgroup meeting https://github.com/WebAssembly/meetings/blob/main/wasi/2023/WASI-09-07.md

PR #560, opened 2023-10-05
Add caveat about support for sync flags

zoraaver:
Since sync flag functionality is entirely dependent on the host platform, add a caveat for each sync flag, warning that functions may return ENOTSUP if an unsupported sync flag is set.


loganek (approved):
makes sense, thanks.

reneleonhardt:
@loganek It's still not possible to improve the documentation or is this PR outdated after two years?

sunfishcode:
Thanks for the ping here!

PR #575, opened 2023-12-16
Add wasi-observe proposal at Phase 0

chrisdickinson:
Hi! This adds a phase 0 proposal for a WASI Observe interface. There's an open discussion issue [here](https://github.com/WebAssembly/wasi-observe/issues/10) and a call for participants issue [here](https://github.com/dylibso/wasi-observe/issues/1). If desired, I can transfer the repo over to WebAssembly on request!

(Thanks for taking a look at this!)


pchickey:
Is it OK if we wait for the phase 1 vote, and then just move this to the phase 1 proposals? Phase 0 exists historically because we wanted to show intent to move forward, but we weren't at the point where we had the tools to break down proxy-wasm into many smaller proposals and move those ahead. That stage is pretty well behind us now, and I don't think we have much reason for Phase 0 existing anymore, so I'd prefer to not keep using it, and we can clean up the historical use sometime soon, now that many (all?) the bits and pieces of proxy-wasm are subsumed by other proposals like wasi-http and etc.

chrisdickinson:
Yep! I'll update the PR and leave it open until the meeting. Thanks!

pchickey (approved):

PR #624, opened 2024-10-16
Move `wasi-i2c` to phase 2

Michielvk:
We noticed that `wasi-i2c`'s phase wasn't updated after moving to phase 2 as [discussed during the WASI SG meeting of May 16 2024](https://github.com/WebAssembly/meetings/blob/main/wasi/2024/WASI-05-16.md). This small PR fixes that.


pchickey:
Thanks for fixing this!

PR #630, opened 2024-11-01
Fix typos discovered by codespell

cclauss:
https://pypi.org/project/codespell

% `codespell --ignore-words-list=wit --skip="./legacy/*" --write-changes`

This assumes that we should not modify the files in the `legacy` directory.


pchickey (approved):
Thank you

PR #641, opened 2025-01-30
Unify the format of wit files

oovm:
I wrote a tool that does project-level formatting: https://plugins.jetbrains.com/plugin/24001-wit-idl

Also respects the rules defined in the `.editorconfig` file.

The current code style is a bit confusing.

Most file use a mixture of 2 and 4 spaces, some files use 6 spaces, and [one uses 7 spaces](https://github.com/WebAssembly/WASI/blob/5cf84da666f5f8f35834488720e970c318d35f76/wasip2/filesystem/types.wit#L288-L291).

This PR makes all wit files have a unified code style which defined in `.editorconfig`.

---

Changes will be synchronized to other upstreams later. 

Projects with 0.3.0-draft wit still need some additional development work.


sbc100 (commented):
IIUC these are mirrors of upstream files?

Would it make sense instead to apply these changes upstream first, to avoid have large downstream differences?

ricochet:
Thanks for this! These are just mirrors although I don't have an issue with normalizing what we have here while also updating upstream. For the individual proposal repos, we need to also have this run as a lint like we do with [wit-abi-up](https://github.com/webassembly/wit-abi-up-to-date) action.

vados-cosmonic:
Hey @oovm there have been a bunch of changes to the repo in the meantime -- it doesn't look like this PR is in a place where it will land any time soon so I'm going ahead and closing it. 

That said, I'm definitely happy to re-open if you want to take another stab at the more recent codebase and re-run it through your tooling. These days we do have some conventions around WIT files (4 spaces for indentation, for example), so I'm not sure exactly if that will still line up, but happy to take another look/reconsider these changes then!

PR #674, opened 2025-11-21
Fix typo in preview0 / preview1 wasi_event_t.

wffurr:


yoshuawuyts (changes requested):
One minor nit, but otherwise looks good. Thanks!

PR #826, opened 2025-11-25
Migrate proposals into the WASI monorepo

yoshuawuyts:
This is the first step in bringing all the proposals into a single repository, AKA creating a WASI monorepo. This was done in a similar way to https://github.com/bytecodealliance/wasmtime/pull/11815 using the [`git-filter-repo`](https://htmlpreview.github.io/?https://github.com/newren/git-filter-repo/blob/docs/html/git-filter-repo.html) command.

We've opted to follow a structure similar to `webassembly/spec`, bringing all phase 3 proposals under the `proposals/` sub-directory. In a follow-up PR we will mark these proposals with code owners and do additional cleanup work. For now this PR only brings in the sub-directories (modulo `.github` CI flows).

Thanks!


programmerjake:
this inadvertently closed a bunch of unrelated issues since they happen to share the issue numbers with issues from the original repos that a commit in this PR mentioned as closing.

PR #827, opened 2025-11-25
ci: validate wit definitions

ricochet:
This validates the wit definitions for both 0.2.* and 0.3.*

I have not added the update and publish workflows but will add those in a following PR once we land on the approach (use matrix for filter paths so we're only building what changed).

### DIFF
--- a/.github/actions/install-tools/action.yml
+++ b/.github/actions/install-tools/action.yml
@@ -1,0 +1,40 @@
+name: 'Install WIT Tools'
+description: 'Installs wit-deps and wasm-tools for WIT validation'
+
+inputs:
+  wit-deps-version:
+    description: 'Version of wit-deps to install'
+    required: false
+    default: '0.5.0'
+  wasm-tools-version:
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
After we merge this we should add the following ruleset to the repo via **Settings → Rules → Rulesets → New ruleset → Import a ruleset** so we require all status checks to pass before a PR can be merged:

```json
{
  "name": "Require All Status Checks to Pass",
  "target": "branch",
  "enforcement": "active",
  "conditions": {
    "ref_name": {
      "include": [
        "~DEFAULT_BRANCH"
      ],
      "exclude": []
    }
  },
  "rules": [
    {
      "type": "required_status_checks",
      "parameters": {
        "strict_required_status_checks_policy": true,
        "do_not_enforce_on_create": false,
        "required_status_checks": [
          "Validate WIT"
        ]
      }
    }
  ],
  "bypass_actors": []
}
```

yoshuawuyts (dismissed):
Neatttt, ty!

PR #831, opened 2025-11-26
Move documentation into `docs/` dir

yoshuawuyts:
This PR centralizes all of the WASI documentation files in the `docs/` dir. This includes sources from:

- The project root dir
- The `proposals/` dir
- The main README.md file

The goal with this is make it easier to find what people are looking for. As well as setting us up for further iteration on the documents themselves in the future. We might potentially even deputize specific owners or editors for the docs in the future other than the WASI SG chairs.

~~One notable change in this PR is that it moves the `Contributing.md` file from the project root to the `docs/` directory. But according to GitHub that is a valid place to put it, and will be detected by the GitHub UI.~~

The other notable change is that this moves `Proposals.md` out of the project root. I'm marking this PR as a draft since there are a number of external resources that link to `Proposals.md` on main. I want to file draft PRs against all known references in both the WebAssembly and BytecodeAlliance orgs before marking this PR as ready to review.

Thanks!

PR #832, opened 2025-11-26
Move WASI logo to `assets/` sub-dir

yoshuawuyts:
Sibling to #831, moves the `WASI.png` logo to a dedicated `assets/` sub-dir. This further tidies up the project root, and sets us op to potentially add more assets in the future. For example: I created dedicated Champion teams in the WebAssembly org yesterday, and each of those might want to have an avatar image. I don't know if we'll want to create any (even as a placeholder), but this at least give us the room to.

Just like #831 I'm initially marking this as a draft so I can file draft PRs against all other WebAssembly and Bytecode Alliance repositories to update the paths. Once those are in place I'll mark this as ready to be reviewed. Thanks!


yoshuawuyts:
Scanned all repos, and the only PR to file was: https://github.com/WebAssembly/meetings/pull/1948. This has now been filed, which means this PR is now ready for review.

PR #852, opened 2025-12-16
Specify unlink-file-at in terms of POSIX.

wingo:
Unfortunately only MacOS has the POSIX behavior of returning EPERM when unlinking a directory; Linux returns EISDIR, and Windows returns EACCESS.  Because EACCESS may arise for other reasons but which are outside the purview of WASI, we map all EACCESS to EPERM.

### DIFF
--- a/proposals/filesystem/wit-0.3.0-draft/types.wit
+++ b/proposals/filesystem/wit-0.3.0-draft/types.wit
@@ -595,9 +595,14 @@ interface types {
 
         /// Unlink a filesystem object that is not a directory.
         ///
-        /// Return `error-code::is-directory` if the path refers to a directory.
-        /// Note: This is similar to `unlinkat(fd, path, 0)` in POSIX.
-        @since(version = 0.3.0-rc-2026-02-09)
+        /// This is similar to `unlinkat(fd, path, 0)` in POSIX.
+        ///
+        /// Error returns are as specified by POSIX.
+        ///
+        /// If the filesystem object is a directory, `error-code::access` or
+        /// `error-code::is-directory` may be returned instead of the
+        /// POSIX-specified `error-code::not-permitted`.
+        @since(version = 0.3.0-rc-2025-02-09)
         unlink-file-at: async func(
             /// The relative path to a file to unlink.
             path: string,


badeend:
If all major OS flavors return something different, then clearly POSIX is not the authoritative source here. Additionally, the page you referenced includes the footnote:

> The standard developers reviewed TR 24715-2006 and noted that LSB-conforming implementations may return [EISDIR] instead of [EPERM] when unlinking a directory.

I personally don't see anything wrong with Linux' behavior of returning EISDIR; it is more to-the-point and precisely describes why it failed, instead of a catch-all EPERM.

wingo:
Ah, thank you for that footnote, I missed it!  And I do like Linux's return code!  But if we want one semantics, EISDIR isn't it, as we can't translate results from other platforms without races.

Another option would be to relax WASI to allow but not require `is-directory`.

badeend:
> if we want one semantics

As a general piece of advice: I'd keep away from trying to synthesize a single semantic model for "filesystems" in general. Each OS works differently, and even within a single OS, different filesystem implementations (NTFS, FAT32, EXT4, ..) can produce different results.

> Another option would be to relax WASI to allow but not require `is-directory`

That sounds like a good solution to me. 👍

vados-cosmonic:
Hey @wingo thanks for this contribution and the discussion -- would you be open to making the note even more specific? 

If we're going to note that there are some discrepancies we might as well be specific about which platforms do what to prevent confusion down the line, until this is fixed/not the case.

> Another option would be to relax WASI to allow but not require is-directory.

Are you up for contributing this? This sounds like the right solution (regardless of whether this PR is merged in the meantime). If we go in this direction then the note should actually be more specific (looks like in any case it should be more specific), and it will be up to implementers to follow the semantics (or choose to return `error-code::is-directory`)

wingo:
Thanks for review, updated the text to instead allow `is-directory` and `access` in addition to `not-permitted`.  WDYT @badeend @vados-cosmonic ?

vados-cosmonic:
This looks great to me! If you want to resolve the conflict that would be great, let's get this in ASAP as long as @badeend doesn't have any objections

badeend:
LGTM 👍

wingo:
rebased, ptal !

vados-cosmonic (approved):
LGTM 🚀

PR #867, opened 2026-01-07
lint since

### LINKED ISSUE #865: CI enhancement: lint for since version
For both WASIP2 and WASIP3, we intend for every type, interface, and world to have a `@since` annotation. Given this, we should add automation that lints for this.

ricochet:
Linter for `@since` Annotations in WIT Files:
- Validates `@since` or `@unstable` annotations on top-level declarations (interfaces, worlds, types, records, variants, enums, flags, resources)
- Excludes deps/ directories (vendored code)
- Outputs GitHub Actions annotation format for clickable error links
- Can be run standalone: node .github/scripts/validate-since.js <directory>

I think a better approach would be to have the AST and inspect the annotations directly rather than regexing the source, but this was faster to implement and immediately useful. If we ever run into issues, I'm game to convert this to something more robust.

Added missing `@since` annotations to:
- proposals/filesystem/wit/types.wit - directory-entry, error-code
- proposals/filesystem/wit-0.3.0-draft/types.wit - directory-entry, error-code
- proposals/http/wit-0.3.0-draft/types.wit - 18 declarations (interface, variants, records, types, resources)
- proposals/http/wit-0.3.0-draft/worlds.wit - 4 declarations (2 worlds, 2 interfaces)

When a PR changes WIT files, it will:
1. Validate WIT syntax (existing)
2. Validate wasm encoding (existing)
3. Validate `@since` annotations (new) - fails CI if any top-level declaration is missing `@since` or `@unstable`

Example error output:
`::error file=proposals/http/wit/types.wit,line=5::Missing @since annotation for interface 'types'`

Fixes #865


badeend:
> I think a better approach would be to have the AST and inspect the annotations directly rather than regexing the source, but this was faster to implement and immediately useful. If we ever run into issues, I'm game to convert this to something more robust.

Well, as long as it works: 👍

yoshuawuyts:
@ricochet can you rebase the commits here to put `@since` in backticks? Right now we're mentioning a GitHub user, which will ping them anytime this repo is rebased and the like.

Screenshot 2026-01-07 at 18 35 25

PR #869, opened 2026-01-10
Remove double async'ness from `read-directory`.

badeend:
The method also  returns a `future` so all concurrency can be encapsulated through that.


lukewagner:
Makes sense to me, based on the general rule we discussed in #870.

alexcrichton (approved):

ricochet (approved):

PR #877, opened 2026-02-02
ci: upload wit tar for releases

### LINKED ISSUE #875: Self-contained WIT artifacts as directories
One thing I think would be nice to have as part of the WASI release process would be an artifact that's just a `wit` directory with all of the WIT files contained within. This would enable using WITs without also requiring a dependency on `wkg` which I feel can be useful from a "lean tooling" perspective and also helps transition to a world where WITs are fully managed by a tool eventually. In the long run though I think there'll always be use cases for taking the WITs from here as-is in a simple-ish fashion (e.g. `wkg`-less) and integrating them into a project.

ricochet:
Add new job to create and upload a wasi-wit-VERSION.tar.gz artifact
to GitHub releases containing consolidated WIT files for each proposal.

Note each proposal will include it's deps. These are duplicated across
proposals to keep each proposal self-contained.

NOTE: rc tarballs will fail until we have an updated wkg see [#2434](https://github.com/bytecodealliance/wasm-tools/issues/2434)

Closes #875

### DIFF
--- a/.github/workflows/publish-proposal.yml
+++ b/.github/workflows/publish-proposal.yml
@@ -25,7 +25,7 @@ on:
         description: 'Version of wkg to install'
         required: false
         type: string
-        default: '0.13.0'
+        default: '0.14.0'
 
 jobs:
   publish:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -209,7 +209,7 @@ jobs:
 
       - name: Install wkg and wasm-tools
         run: |
-          cargo binstall -y wkg@0.13.0
+          cargo binstall -y wkg@0.14.0
           cargo binstall -y wasm-tools@1.244.0
 
       - name: Create consolidated WIT directory


alexcrichton:
Looking over the 0.2.10 release to double-check on this, the current structure is:

```
wasi-wit-0.2.10
wasi-wit-0.2.10/http
wasi-wit-0.2.10/http/http.wit
wasi-wit-0.2.10/http/deps
wasi-wit-0.2.10/http/deps/clocks.wit
wasi-wit-0.2.10/http/deps/cli.wit
wasi-wit-0.2.10/http/deps/random.wit
wasi-wit-0.2.10/http/deps/io.wit
wasi-wit-0.2.10/random
wasi-wit-0.2.10/random/random.wit
wasi-wit-0.2.10/io
wasi-wit-0.2.10/io/io.wit
wasi-wit-0.2.10/clocks
wasi-wit-0.2.10/clocks/clocks.wit
wasi-wit-0.2.10/clocks/deps
wasi-wit-0.2.10/clocks/deps/io.wit
wasi-wit-0.2.10/sockets
wasi-wit-0.2.10/sockets/deps
wasi-wit-0.2.10/sockets/deps/clocks.wit
wasi-wit-0.2.10/sockets/deps/io.wit
wasi-wit-0.2.10/sockets/sockets.wit
wasi-wit-0.2.10/filesystem
wasi-wit-0.2.10/filesystem/filesystem.wit
wasi-wit-0.2.10/filesystem/deps
wasi-wit-0.2.10/filesystem/deps/clocks.wit
wasi-wit-0.2.10/filesystem/deps/io.wit
wasi-wit-0.2.10/cli
wasi-wit-0.2.10/cli/cli.wit
wasi-wit-0.2.10/cli/deps
wasi-wit-0.2.10/cli/deps/filesystem.wit
wasi-wit-0.2.10/cli/deps/clocks.wit
wasi-wit-0.2.10/cli/deps/sockets.wit
wasi-wit-0.2.10/cli/deps/random.wit
wasi-wit-0.2.10/cli/deps/io.wit
```

would it be possible to change that instead to:

```
wasi-wit-0.2.10
wasi-wit-0.2.10/http.wit
wasi-wit-0.2.10/random.wit
wasi-wit-0.2.10/io.wit
wasi-wit-0.2.10/clocks.wit
wasi-wit-0.2.10/sockets.wit
wasi-wit-0.2.10/filesystem.wit
wasi-wit-0.2.10/cli.wit
```

Not urgent by any means, but that should make consumption a bit easier

ricochet:
One key property was to keep each proposal self-contained, which required including the deps otherwise consumers will need another tool to work as a deps-solver. The idea being folks could ignore/remove the deps folder if that's not required for their use. The deps dir per proposal is what forces this into a non-flat structure.

If we want to remove that property of the artifact for some other ergonomic need, I'd like to understand it a little more. I like that individual proposals can be dropped in with existing tooling and work today and this continues the modularity goal of wasi proposals vs treating them as a monolith.

alexcrichton:
Ok that makes sense yeah. I was thinking we'd just splat the tarball's contents into a `deps/*.wit` location and it'd be easier than rooting around folders, but on second thought I like the deps being "auto trimmed" in this representation. Disregard me!

PR #884, opened 2026-02-15
sockets: Update documentation

badeend:
Update sockets documentation:
- Limit the prose to about 80 characters per line.
- Describe POSIX equivalents of the error codes.
- Removed outdated `stream`/`error-context` info from:
    - `resolve-addresses`
    - `tcp-socket::receive`
- Describe shared ownership semantics of sockets and their streams/futures that we landed on in https://github.com/WebAssembly/WASI/pull/870
- Capture the "lessons learned" so far:
    - TCP & UDP `connect` may update the socket's local address. See https://github.com/bytecodealliance/wasmtime/issues/12589
    - Recommend to always set the SO_REUSEADDR flag on TCP `bind` (except on Windows). See https://github.com/bytecodealliance/wasmtime/issues/12402
    - send/receive buffer sizes may be dynamically adjusted by the OS. See https://github.com/bytecodealliance/wasmtime/issues/12561
    - Windows does not perform an implicit bind on UDP `send`. See https://github.com/bytecodealliance/wasmtime/pull/12225

### DIFF
--- a/proposals/filesystem/wit-0.3.0-draft/types.wit
+++ b/proposals/filesystem/wit-0.3.0-draft/types.wit
@@ -595,9 +595,14 @@ interface types {
 
         /// Unlink a filesystem object that is not a directory.
         ///
-        /// Return `error-code::is-directory` if the path refers to a directory.
-        /// Note: This is similar to `unlinkat(fd, path, 0)` in POSIX.
-        @since(version = 0.3.0-rc-2026-02-09)
+        /// This is similar to `unlinkat(fd, path, 0)` in POSIX.
+        ///
+        /// Error returns are as specified by POSIX.
+        ///
+        /// If the filesystem object is a directory, `error-code::access` or
+        /// `error-code::is-directory` may be returned instead of the
+        /// POSIX-specified `error-code::not-permitted`.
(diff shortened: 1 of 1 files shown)


proposals/sockets/wit-0.3.0-draft/types.wit
```diff
@@ -709,11 +782,14 @@ interface types {
...
+        /// Kernel buffer space reserved for sending/receiving on this socket.
+        /// Implementations usually treat this as a cap the buffer can grow to,
+        /// rather than allocating the full amount immediately.
         ///
         /// If the provided value is 0, an `invalid-argument` error is returned.
-        /// Any other value will never cause an error, but it might be silently clamped and/or rounded.
-        /// I.e. after setting a value, reading the same setting back may return a different value.
+        /// Any other value will never cause an error, but it might be silently
+        /// clamped and/or rounded. I.e. after setting a value, reading the same
+        /// setting back may return a different value.
```
ricochet: ditto
```suggestion
        /// All other values are accepted without error, but may be
        /// clamped or rounded. As a result, the value read back from
        /// this setting may differ from the value that was set.
```

badeend:
@ricochet Thanks for the textual improvements. I've accepted them all.

PR #920, opened 2026-05-28
Document concurrent instance reuse in `wasi:http/service`

### LINKED ISSUE #918: wasi-http 0.3: `request` and `response` should have a tracing identifier
In wasi-http 0.3, assumptions about instance reuse are changing: see #898. If other discussions exist where this has been discussed, please tag those in.

When instance reuse is considered the default, wasi-http embeddings can no longer depend on considering all output associated with a given instance to be rooted from a single `incoming-request`. Output includes prints to stdout and stderr, structured logs over other interfaces, outgoing requests (http and otherwise), database transactions, etc. No longer being able to observe what outputs are associated with what inbound request can be a significant user experience degradation (e.g. https://github.com/bytecodealliance/wasmtime/pull/13404#discussion_r3262826113) and could potentially block adoption of instance reuse if unsolved.

The essential problem now is that a p3 instance exporting a wasi-http `handler` has no way to assign an identifier to an owned `request` that corresponds to anything the callee considered the identity of that `request`. The outside world - whether the host calling into the `handler`, or another component doing the same - has a mechanism to uniquely identify each `request` it creates, and so does the guest (as a index into the guest instance's own table), but (by the component model's design) neither has a mechanism to map their own identifier to another's identifier.

The simplest possible thing we can do to make the same identifier legible to both sides of an interface is to give each `request` a method to retrieve an associated UUID. UUIDs are 128 bit numbers which commonly are presented as a string, but the representation as just 128 bits (morally a `u128` which does not exist in the wit type system) is more compact and does not privilege any of the evolving mechanisms for string presentation.

So, I propose we add to the wasi http types wit:

```wit
record uuid { upper: u64, lower: u64 }

resource request {
    /// Get the universally unique identifier associated with the request. For a given request, this always returns
    /// the same value, and implementations must make best effort for that identifier to be universally unique.
    get-id: func() -> uuid;
}
```

The more difficult part of the design is that `response`s have the same problem. In the HTTP protocol a Response always corresponds with exactly one Request, but the wit interface (correctly, in my opinion) does not retain that property - the `response` resource can be constructed without any associated `request`. As a strawman, I'd suggest that we make a uuid an optional property of the response, to be provided in the constructor or set at most once, so that the getter can indicate that either no request is yet associated with this response, or that the same request is associated with the response for the entire lifetime of the response. If no uuid is ever set, it is not considered an error.

```wit
resource response {
    /// added a constructor argument for optional uuid, which should only be set to the identifier from an
    /// associated `request` if it is avaiblable.
    new: static func(
      headers: headers,
      contents: option<stream<u8>>,
      trailers: future<result<option<trailers>, error-code>>,
      id: option<uuid>,
    ) -> tuple<response, future<result<_, error-code>>>;

    /// Get the universally unique identifier associated with the response, if one has been assigned. If this returns
    /// some, it will return that same uuid for the entire lifetime of the response.
    get-id: func() -> option<uuid>;
    /// Set the universally unique identifier associated with the response, if one has not yet been set. If one has been
    /// set previously (including providing it in the constructor), returns an error. This identifier should only be set to the
    /// identifier from an associated `request`.
    set-id: func(id: uuid) -> result;
}
```

Given its impact on the usability of the exciting new instance reuse features, I believe we should try to get this addition into wasip3 before it launches if possible.

### LINKED ISSUE #919: wasi-http 0.3: instance reuse means middleware world cannot implement CDN-Loop
The wasi-http 0.3 middleware world, as well as the work done to make instance reuse the default, are both features I'm strongly in favor of. The trouble comes with the intersection of those features, which I think breaks some assumptions. @cfallin first brought this issue up in [#918](https://github.com/WebAssembly/WASI/issues/918#issuecomment-4493009691).

[CDN-Loop](https://datatracker.ietf.org/doc/html/rfc8586) is a header that we expect many wasi-http embedders to require in their implementations of wasi-http middleware worlds. In short, the implementation will read the CDN-Loop header out of the incoming request, parse out the values specific to their CDN, possibly choose to block the request if those values exceed some limit, and otherwise the implementation will hand the request to a wasi-http middleware world's exported `handler`. Then, the implementation of the wasi-http middleware's imported `handler` needs to take the request and add a CDN-Loop header with all of the values from the inciting request's header, plus an added (on first hop) or modified (when appending a hop) value specific to the CDN. This is straightforward to implement in wasi-http 0.2, and indeed many implementations do so today.

However, with instance reuse, once there are multiple simultaneous requests passed to the exported `handler`, it is no longer possible for an implementation to associate a call of middleware's imported `handler` with the inciting request of middleware's exported `handler`. This ambiguity makes impossible to implement CDN-Loop correctly, and if implemented incorrectly CDN-Loop could end up either incorrectly blocking benign requests, or failing to block requests that should be blocked, neither of which a CDN can tolerate.

Inside any correct middleware component, with instance reuse or not, there is an association between an incoming request and the outgoing request, in order to turn the outgoing request's response to the incoming request's response. Instance reuse, and the current shape of the middleware wit world, makes it impossible for the implementation of the middleware world to maintain that association. Host magic to solve this doesn't exist, and even if it did, we want this exact sort of thing to be implementable with virtualization anyway.

CDN-Loop is just one example in the broader category of problems caused by loosing association between the inbound request and outbound request in a middleware or proxy. This problem presents in other ways specific to the embeddings that @cfallin and I maintain for F5, as well, but CDN-Loop is a representative example that exists across many vendors, so I figured its the best way to discuss it in the standard.

alexcrichton:
This commit adds a documentation block to the `wasi:http/service` world that's the result of discussions in today's WASI subgroup meeting, #918, and #919. The documentation here indicates a few properties around concurrent instance reuse, namely:

* Guests should expect to be concurrently reused. If this is not acceptable then guests should disable reuse through backpressure mechanisms.

* Guests should ensure that outgoing HTTP requests are done within the context of the original root task, or the original invocation of `handler`. If guests can't implement this then they're expected to use backpressure mechanisms instead.

The overall conclusion of the various discussions that have happened over the past week or so is that it's not possible to provide the attribute behavior some hosts require exclusively through either in-guest interactions or pure-host interactions. Instead being able to reliably provide accurate attribution requires specifically outlining guidelines for both guests and hosts. The mechanism intended to be used for attribution is then:

* For hosts such as Wasmtime calls to imports will be able to reflect on the current async call stack, notably the root export task. Hosts can then use this to correlate an identifier provided when a root task is created, providing a strong link from the import call to whatever export initiated it. This is expected to work as-is for some guest languages, such as Rust, but this is also not a bulletproof solution. Guests can always internally call imports from any currently-running task, and additionally languages such as JS and Go at this time are structured in such a way where imports are not always naturally called from the originating export task.

* To enable languages like JS and Go to be able to leverage concurrent reuse while providing accurate attribution of outgoing requests to incoming requests, the plan is to eventually add and specify intrinsics on the component model level to, in a scoped fashion, mutate the async call stack. For example a guest would be able to say "I'm about to do some work for this component model task". When Wasmtime reflects on the async call graph at that time it'd see this and understand that the attribution needs to be slightly adjusted to what it would otherwise by default be.

The goal with these addition is to enable hosts to be able to reliably expect to attribute outgoing requests to incoming requests (via async call graph inspection), provide guests the ability to work today (either naturally or disabling concurrent reuse via backpressure), and ensure that well-behaved and idiomatic guests can work reliably with concurrent reuse in the long-run (via component model intrinsics to massage the async backtrace that hosts operate with). When this is all combined it's expected to resolve the concerns of #918 and #919 with idiomatic bindings in guests and hosts alike.

Closes #918
Closes #919


lukewagner (dismissed):
Looks good from my perspective

badeend (dismissed):
Agree with these changes. 

I also think that using the CM async callstack to correlate exports with imports is the best (or: least bad) solution put forward so far to solve #918 & #919.

proposals/http/wit-0.3.0-draft/worlds.wit
```diff
@@ -31,6 +31,32 @@ world service {
   ///
   /// This may also be used to receive synthesized or forwarded requests from
   /// another component.
+  ///
+  /// # Concurrent instance reuse
+  ///
+  /// Hosts may concurrently invoke this `handler` interface multiple times,
+  /// even when previous requests have yet to finish processing. This means that
+  /// the guest, if it blocks, may be required to service multiple requests at
```
tschneidereit: "if it blocks" might not be particularly clear
```suggestion
  /// the guest, if it becomes idle while being blocked on I/O, may be required to service multiple requests at
```

proposals/http/wit-0.3.0-draft/worlds.wit
```diff
@@ -31,6 +31,32 @@ world service {
...
+  /// for example, to correctly implement the CDN-Loop header automatically and
+  /// ensure that outgoing requests correctly build on incoming requests. In
+  /// general this is not a hard requirement that embedders can rely on, but
+  /// well-behaved guests are expected to provide this guarantee. Hosts which
+  /// require a hard guarantee regardless of guest will not invoke `handler`
+  /// more than once concurrently. Guests which export `handler` are expected to
+  /// ensure that calls to imported interfaces happen on the same component
+  /// model task that the `handler` was originally invoked on. If guests cannot
+  /// provide this guarantee then they're expected to opt-out of concurrent
+  /// through the component model's backpressure mechanisms.
```
tschneidereit: I took a stab at restructuring this to spell out host and guest expectations clearly, with the reasoning behind them extracted in a way that I hope to make it easier to understand what this is about:
```suggestion
  /// In some contexts, correct attribution of resource usage and I/O operations
  /// is an important requirement. For example, correct implementation of the
  /// CDN-Loop header in the context of a proxy server requires that outgoing
  /// HTTP requests can be attributed to the incoming HTTP request that induced
  /// them.
  ///
  /// Hosts with this requirement must associate work happening on the same
  /// component model task that the `handler` was originally invoked on with the
  /// same incoming HTTP request.
  ///
  /// Conversely, guests must ensure that calls to imported interfaces happen on
  /// the same component model task that the `handler` was originally invoked
  /// on. Guest runtime implementations that do not provide this guarantee are
  /// expected to opt-out of concurrent through the component model's
  /// backpressure mechanisms.
  ///
  /// Note that hosts can't rely on this guarantee being upheld by all guests,
  /// and shouldn't treat it as part of their security model. Instead, the
  /// combination of the above host and guest requirements enable hosts to
  /// enforce correctness properties on behalf of the overall system a guest is
  /// part of.
```

tschneidereit (commented):
I added suggestions for fairly meaningful changes to the comments, which I hope make things a bit clearer.

tschneidereit (approved):

proposals/http/wit-0.3.0-draft/worlds.wit
```diff
@@ -31,6 +31,38 @@ world service {
...
+  /// is an important requirement. For example, correct implementation of the
+  /// CDN-Loop header in the context of a proxy server requires that outgoing
+  /// HTTP requests can be attributed to the incoming HTTP request that induced
+  /// them.
+  ///
+  /// Hosts with this requirement must associate work happening on the same
+  /// component model task that the `handler` was originally invoked on with the
+  /// same incoming HTTP request.
+  ///
+  /// Conversely, guests must ensure that calls to imported interfaces happen on
```
dicej:
```suggestion
  /// Conversely, guests should ensure that calls to imported interfaces happen on
```

I think we should soften this to "should" for two reasons:

- Saying "must" implies that hosts can rely on it, which they can't.  It will always be possible to write a guest which neither links import calls to export calls in the way the host expects nor applies backpressure.  I see you noted that in the last paragraph, but the softer language here will help to underscore it.
- The component author may simply not care about preserving task attribution, nor the benefits it may provide, nor the hazards it might avoid.  If they write a component that e.g. sends all work to a single background task with no attempt to preserve call stack attribution, it's still a valid component, and the host has no reliable means to detect the "misattribution" and reject or trap as a consequence.  The consequence must be limited to the host e.g. spuriously refusing to send the request due to a CDN loop counter being exceeded sooner than it otherwise would have, or inaccurate host-side tracing, or similar.

proposals/http/wit-0.3.0-draft/worlds.wit
```diff
@@ -31,6 +31,38 @@ world service {
...
+  /// the same component model task that the `handler` was originally invoked
+  /// on. Guest runtime implementations that do not provide this guarantee are
+  /// expected to opt-out of concurrent through the component model's
```
dicej:
```suggestion
  /// encouraged to opt-out of concurrent through the component model's
```

See my earlier comment for rationale.

pchickey (dismissed):

dicej (approved):

proposals/http/wit-0.3.0-draft/worlds.wit
```diff
@@ -31,6 +31,38 @@ world service {
...
+  /// them.
+  ///
+  /// Hosts with this requirement must associate work happening on the same
+  /// component model task that the `handler` was originally invoked on with the
+  /// same incoming HTTP request.
+  ///
+  /// Conversely, guests should ensure that calls to imported interfaces happen on
+  /// the same component model task that the `handler` was originally invoked
+  /// on. Guest runtime implementations that do not provide this guarantee are
+  /// encouraged to opt-out of concurrent through the component model's
```
dicej:
```suggestion
  /// encouraged to opt-out of concurrent reuse through the component model's
```

ricochet (dismissed):

yoshuawuyts (approved):
This looks good to me! - We discussed this change sync during the WASI meeting two weeks ago, and this matches what I expected it to say. Thank you for putting this together!

PR #922, opened 2026-06-02
feat(cli): exit-with-code function stable

ricochet:
Remove the feature gate for stabilized feature `exit-with-code function`


ricochet:
I'm going to add some release automation for promoting stabilized features

PR #923, opened 2026-06-02
stabilize next

ricochet:
Add release automation for stabilizing features.

When a feature is voted as stabilized, add the `@stablize-next` annotation above the feature gate, and the release automation will remove the gate after rev'ing to the next version as part of the automated release.

The second commit adds `@stablize-next` to `cli-exit-with-code`.


yoshuawuyts (approved):
Oh I like this even better than what we discussed. Neat!

PR #926, opened 2026-06-04
Allow re-casing when serializing for transmission

cceckman-at-fastly:
HPACK (RFC 7541) provides a mechanism for compressing HTTP headers; it is used in HTTP/2 and HTTP/3. One of the ways it accomlishes compression is via a static table of common values, listed in Appendix A. Of note, all of the field names in this table are provided in lower case.

The restriction of preserving field name case when serializing for transmission implies that an implementation of wasi-http may not always be able to use the most efficient encoding of the field name (from the static table), even when the semantics are the same.

This commit relaxes the requirement to preserve case _when serializing for transmission_. This allows an implementation to use the most efficient encoding available when the result is semantically equivalent, e.g. to use HPACK ignoring case differences.


pchickey (dismissed):

cceckman-at-fastly:
Revised as the files moved around; and included a note in the comment as to _why_ case matters.

lukewagner (approved):
(Sorry for the delay!  This fell off my radar) This change makes sense to me and I believe it's necessary even independently of HPACK optimizations: some native HTTP clients will just do this in any case.

lukewagner (approved):

PR #956, opened 2026-09-01
Remove myself from WASI HTTP champions.

PiotrSikora:
I wasn't active since the initial proposal, so this is long overdue.

While there, remove Proxy-Wasm from stage 0, since all subcomponents
advanced as separate proposals.


devigned (approved):
lgtm

lukewagner:
Merging after WASI-09-03 vote (additional champions incoming in a separate PR).

